PR #6, opened 2023-03-14
fix(llama): use ggml_type to build on Windows

philpax:
Pretty simple fix. Encountered this in my attempt as well - `ggml_type` can be different between operating systems thanks to the loosy-goosy C integer rules 🙃


setzer22:
Good catch! Thanks a lot :)

PR #7, opened 2023-03-14
fix(main): use args.model_path

### LINKED ISSUE #2: hardcoded model_path, ignoring -m
It appears that the model path CLI argument is being ignored entirely and that the path to the model is hardcoded.

philpax:
Another simple fix - removes the hardcoded model path.

Fixes #2.


setzer22:
My bad 🤦  Thanks a lot for the PR!

PR #10, opened 2023-03-15
Librarification

philpax:
This is a very rough split of the CLI application into a library and application. It includes #8 because that was on my branch when I did this - we can figure out what to do with it when it's time.

The main differences are
- conversion of `println`s to `log`s
- removal of `print`s
- adding a callback function for arrival of new tokens

Turns out this is basically entirely sufficient for first-pass use as a library. [Here's proof.](https://github.com/philpax/llamacord)


mwbryant (approved):
Everything LGTM with some very minor nits.  I've tested it and with a fixed seed it gives the same results and performance as the original.  Needing to set the flags (which also required Rust 1.68) for the build script is still a pain but I'm sure there's a solution to that.  

Seems to be resilient to all the crazy things I tried, inference can be called multiple times without issue.  

In the future it would be nice to have inference with a continuous state between calls (i.e. an interactive mode)

llama-rs/src/llama.rs
```diff
@@ -558,14 +577,10 @@ impl LlamaModel {
                 }
 
                 n_tensors += 1;
-                if n_tensors % 8 == 0 {
```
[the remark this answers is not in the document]
setzer22: Agreed. Maybe we could do the same thing that was done with the token generation, and provide a callback argument that will be called every time this was printed, so the calling side (in this case `llama-rs-cli`) can be a bit more opinionated and print the loading bar there.

ggml-raw/build.rs
```diff
@@ -10,13 +9,64 @@ fn main() {
...
-    build.flag("-mavx");
-    build.flag("-mfma");
-    build.flag("-mf16c");
+    // This is a very basic heuristic for applying compile flags.
+    // Feel free to update this to fit your operating system.
+    let target_arch = std::env::var("CARGO_CFG_TARGET_ARCH").unwrap();
+    let target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap();
+    let is_release = std::env::var("PROFILE").unwrap() == "release";
+
+    let supported_features: HashSet<_> = std::env::var("CARGO_CFG_TARGET_FEATURE")
```
setzer22, line 18: If this new version includes the changes from #8, we need to adjust the README to include instructions.

I think it's okay to merge this for the time being, but I would say in the long run:

- We use something similar to llama.cpp's autodetection (shelling out to commands to autodetect CPU features).
- We allow overriding those heuristics via some env variable, where the user specifies exactly what they want. Useful for cross-compilation.

This makes the default simple for the vast majority of use cases, and if well documented, it wouldn't make things much more complex for people attempting cross-compilation.

setzer22 (commented):
This was quick! :smile:  Thanks a lot for the changes.

Not much to add on top of @mwbryant's review, but I left a couple additional comments.

llama-rs/src/llama.rs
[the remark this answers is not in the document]
setzer22: Computing the amount of tensors shouldn't be hard. The code already creates each tensor before loading them, so we definitely could check how many there are.

There's also an already computed `ctx_size` and `total_size` variables (the latter is updated as tensors are loaded). But I'm not sure if one will end up accurately matching the other.

llama-cli/src/main.rs
```diff
@@ -35,9 +35,12 @@ fn main() {
...
     };
 
-    let (model, vocab) = llama::LlamaModel::load(&args.model_path, args.num_ctx_tokens as i32)
+    let (model, vocab) = llama_rs::Model::load(&args.model_path, args.num_ctx_tokens as i32)
         .expect("Could not load model");
 
     let mut rng = thread_rng();
-    model.inference_with_prompt(&vocab, &inference_params, &prompt, &mut rng);
+    model.inference_with_prompt(&vocab, &inference_params, &prompt, &mut rng, |t| {
+        print!("{t}")
```
setzer22, line 43: I just noticed something: You need to call `std::io::stdouot().flush()` here.

philpax:
Thanks for the feedback, everyone! Will action within the next few hours so we can get this merged 👍

ggml-raw/build.rs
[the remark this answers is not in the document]
philpax: The MSRV is whatever it currently is [we should check, `cargo-msrv` does well at this] but only 1.68.0 supports f16c in `target-features`. My changes to #8 sidestep this by checking for f16c on the host by default.

llama-rs/src/llama.rs
```diff
@@ -575,12 +590,13 @@ impl LlamaModel {
         Ok((model, vocab))
     }
 
-    pub fn inference_with_prompt(
+    pub fn inference_with_prompt<'a>(
```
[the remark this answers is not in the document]
philpax: Good catch! Fixing.

ggml-raw/build.rs
philpax: We can use the improved #8, or I can remove #8 from this PR. Either one is fine by me.

philpax:
The `llama-rs` crate now has no logging of its own; all progress is reported through the callback. Pretty happy with how that turned out!

I've also switched over to `thiserror`. I've tried to remain consistent with the existing names, but I'd like to agree on #17 before we have any major users.

setzer22:
> As an aside: should the CLI be llama-cli or llama-rs-cli

I was originally thinkinc llama-rs-cli. But after seeing the name, I really like `llama-cli`, especially if it ends up in someone's path via cargo install, it's easier to type. So if you prefer that one I'm on board as well!

llama-cli/src/main.rs
```diff
@@ -0,0 +1,102 @@
+use std::io::Write;
+
+use cli_args::CLI_ARGS;
+use llama_rs::InferenceParams;
+use rand::thread_rng;
+
+mod cli_args;
+
+fn main() {
+    env_logger::init();
```
setzer22, line 10: We should make the `info` log level the default for the CLI app. The output doesn't look too noisy and model loading times are large enough that most users are going to want to see some progress indication, especially important for the first-time experience.
```suggestion
    env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("info")).init();
```

Doing this allows setting the `RUST_LOG` environment variable to a higher level (like `warn` or `error`) so users who don't want to see the logs still get the choice to do so.

llama-rs/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+mod ggml;
+mod llama;
+
+pub use llama::{
+    GptVocab as Vocab, InferenceParams, LlamaHyperParams as HyperParams, LlamaModel as Model,
+    LoadProgress, OutputToken,
```
setzer22, line 6: We should just do the renaming we discussed here :+1:

llama-rs/src/llama.rs
```diff
@@ -105,32 +124,108 @@ fn llama_n_parts(size: i32) -> i32 {
     }
 }
 
+/// Each variant represents a step within the process of loading the model.
+/// These can be used to report progress to the user.
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
+pub enum LoadProgress<'a> {
```
setzer22, line 130: A very clean solution! :smile:  Having this will be especially important if we end up implementing parallel loading as described on #15. Because relying on the order of print messages for the little dots breaks the moment you start multithreading.

setzer22 (commented):
Amazing work, thanks a lot again for this!

Left a few minor comments, should be quick enough to address :+1:

llama-rs/src/lib.rs
philpax: Sounds good. Should I do `Vocab` -> `Vocabulary` and `Params` -> `Parameters`?

setzer22: Yup, sounds good

PR #12, opened 2023-03-15
fix some clippy lints

CheatCod:
Open the project with clippy default lints, noticed clippy isn't happy, fixed some of the clippy lints

### DIFF
--- a/llama-rs/src/llama.rs
+++ b/llama-rs/src/llama.rs
@@ -426,41 +426,31 @@ impl LlamaModel {
                 } else if tensor_name.contains("layers") {
                     if tensor_name.contains("attention.wo.weight") {
                         0
-                    } else if tensor_name.contains("feed_forward.w2.weight") {
-                        0
                     } else {
-                        1
+                        i32::from(!tensor_name.contains("feed_forward.w2.weight"))
                     }
-                } else if tensor_name.contains("output") {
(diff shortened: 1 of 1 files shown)


llama-rs/src/llama.rs
```diff
@@ -426,41 +426,31 @@ impl LlamaModel {
...
-                        0
                     } else {
-                        1
+                        i32::from(!tensor_name.contains("feed_forward.w2.weight"))
                     }
-                } else if tensor_name.contains("output") {
-                    1
                 } else {
-                    0
+                    i32::from(tensor_name.contains("output"))
```
mwbryant, line 433: I don't know if this is the right fix for this, honestly split_type should be an enum and the comment already tells us the 2 variants should be Columns and Rows

mwbryant (approved):
LGTM, a larger refactoring is probably needed to clean up more of the Cisms.  I'll look into that after #10 merges

philpax:
Once all the other pending PRs are fixed, I reckon we should go through and fix all the warnings.

setzer22:
Hey! Sorry we didn't merge this sooner. Currently this branch has drifted a lot from `main` and CI is enforcing that there are no errors reported by clippy.

Regardless, thank you very much for the PR :smile:

PR #14, opened 2023-03-15
Implementation of prompt caching

setzer22:
This was simpler than I expected, so time to sneak in a little PR :)

This implements the same feature described in https://github.com/ggerganov/llama.cpp/issues/64

Basically, this adds an API to access the memory tensors in the model, and adds a couple functions to save them and load them from disk.

I also added `--cache-prompt` and `--restore-prompt` flags for `llama-cli`:

- When running with `--cache-prompt` mode, no inference is run, and the program dumps the memory contents into a file after feeding it the prompt.
- When running with `--restore-prompt`, the contents of the memory are read from the disk at the given path, and the prompt you feed the system is concatenated, right after the cached prompt.

Note this PR builds on top of #10, so the diff will not look correct until that is merged. You have to filter to only look at the commits from this PR when you go to the "Files changed" tab.

I have tested the changes, but since this adds a non-trivial change, I'd like to make sure I didn't break anything before merging. Anyone interested please pull the branch and test it before we commit. Here's some minimal instructions.

1. Run the model with an incomplete prompt, you can use the example below as an example:
```shell
RUSTFLAGS='-C target-feature=+avx2,+fma,+f16c'  cargo run --release -- -m /data/Llama/LLaMA/7B/ggml-model-q4_0.bin -f <path_to_prompt_text_file> --cache-prompt <path_to_cache_file>
```
```
The following text is a transcript for a conversation between a human user and The Assistant. The Assistant is a smart, reliable, caring, confident chatbot which is based on advanced AI technology and is capable of replying to the user's messages in a truthful and understanding way.

The transcript consists of an exchange of messages. User messages begin with the word USER, while The Assistant's messages start by using the word ASSISTANT.

=== BEGIN TRANSCRIPT ===
USER: Explain what a game engine is like a 5 year old
ASSISTANT:
```
2. Run the command again, this time passing the `--restore-cache` flag and a smaller prompt
```
RUSTFLAGS='-C target-feature=+avx2,+fma,+f16c'  cargo run --release -- -m /data/Llama/LLaMA/7B/ggml-model-q4_0.bin -p "A game engine" --restore-prompt <path_to_cache_file> 
```

The observed behavior should be that the second time, the system starts predicting just where it left off, and you only pay for the time it takes to parse the extra prompt :)


llama-cli/src/main.rs
```diff
@@ -35,12 +37,51 @@ fn main() {
         std::process::exit(1);
     };
 
-    let (model, vocab) = llama_rs::Model::load(&args.model_path, args.num_ctx_tokens as i32)
+    let (mut model, vocab) = llama_rs::Model::load(&args.model_path, args.num_ctx_tokens as i32)
         .expect("Could not load model");
 
+    if let Some(restore_path) = &args.restore_prompt {
+        let memory = ModelMemory::load_compressed(restore_path);
```
setzer22, line 44: You can load a stored memory file from disk like this. I chose to add a simple API with `bincode`, but the struct implements `De`/`Serialize` so you can do your own serialization instead.

llama-cli/src/main.rs
```diff
@@ -35,12 +37,51 @@ fn main() {
...
+            print!("{t}");
+            std::io::stdout().flush().unwrap();
+        },
+    );
+
+    if let Some(cache_path) = &args.cache_prompt {
+        // SAFETY: no other model functions used inside the block
+        unsafe {
+            let memory = model.get_memory();
+            match memory.write_compressed(cache_path) {
```
setzer22, line 72: And this is how you access tensor memory and write it to a file. 

This returns a reference to tensor memory inside the inner context, so it's not safe to access that and then keep using the `model` object because that might mutate memory underneath and cause UB. I just decided to mark the function `unsafe` out of caution, but there's a whole underlying safety issue here that the `ggml` bindings allow mutating via a shared ref `&`. (more on that below)

llama-rs/src/llama.rs
```diff
@@ -577,7 +611,12 @@ impl LlamaModel {
                 }
 
                 n_tensors += 1;
+                if n_tensors % 8 == 0 {
+                    print!(".");
+                    std::io::stdout().flush();
+                }
             }
+            println!(".");
```
setzer22: Brought this back temporarily because I was going crazy without it :smile:  I'll just end up rebasing on whatever we agree for as part of #10.

llama-rs/src/llama.rs
```diff
@@ -591,54 +630,59 @@ impl LlamaModel {
...
         params: &InferenceParams,
         prompt: &str,
         rng: &mut impl rand::Rng,
+        stop_after_prompt: bool,
         callback: impl Fn(OutputToken<'a>),
     ) {
-        let embd_inp = self.tokenize(vocab, prompt, true);
-        let mut logits = Vec::new();
+        let beginning_of_sentence = self.n_past == 0;
+        let embd_inp = self.tokenize(vocab, prompt, beginning_of_sentence);
```
setzer22, line 642: One sneaky issue I found was that we don't want to feed the beginning of sentence token when restoring from memory, because that causes the system to insert that token into the memory and then the system ignores any tokens before it.

llama-rs/src/llama.rs
```diff
@@ -591,54 +630,59 @@ impl LlamaModel {
...
+        // Mem per token is initialized at zero. Will be read the first time
+        // llama_eval is called.
         let mut mem_per_token = 0;
-        let _ = self.llama_eval(
-            params.n_threads,
-            0,
-            &[0, 1, 2, 3],
-            &mut logits,
-            &mut mem_per_token,
-        );
```
setzer22, line 596: This is one important change that *might* (?) cause issues on the larger models. The way `llama_eval` works is that when you pass mem_per_token as zero, it computes the memory and sets it back for you so you can pass it after successive calls.

But this bogus call caused issues when we were resuming from a cached prompt (because it overwrites some memory tokens). So I decided to simply remove the call. This apparently works just fine, and thinking about it, I'm not even sure what the intent of the original code was, but let's make extra sure just in case :thinking:

llama-rs/src/llama.rs
```diff
@@ -591,54 +630,59 @@ impl LlamaModel {
...
-                    n_past,
+                    self.n_past as i32,
                     &embd,
                     &mut logits,
                     &mut mem_per_token,
                 );
             }
 
-            n_past += embd.len() as i32;
+            self.n_past += embd.len();
```
setzer22, line 670: `n_past` is now stored inside the `LlamaModel` object. This allows invoking the generation multiple times. Wen n_past is > 0, inference continues by adding any extra prompt you feed it, and then starts inferring normally.

This could also serve as the basis for a more complex chat-like interaction :)

llama-rs/src/llama.rs
```diff
@@ -591,54 +630,59 @@ impl LlamaModel {
...
 
             if embd_inp.len() <= input_consumed {
                 // out of input, sample next token
+
+                // TODO: The logic is super convoluted already and this only
+                // makes it worse... Refactor!
+                if stop_after_prompt {
+                    break;
+                }
+
```
setzer22, line 681: This was necessary because we don't want to do any inference when caching prompts, just stop right after the prompt has been fed to the system.

Adding this break statement here made me think how much convoluted this logic is already. But we can fix that in a later PR

setzer22 (commented):
Some comments as a self-review (for those new to the repo, I always try to do this. I think it helps others understand the changes :smile:)

setzer22:
Oh, before I forget. I also tried using the [snap](https://docs.rs/snap/latest/snap/) crate for compression. Some quick results:

- Compression ratios for the prompt I'm sharing above look good. A cached prompt is around ~500MB, but only 250MB when compressed. However, this compression ratio may just be an artifact from having a ton of zeroes at the end of the (unfilled) memory.
- Loading speed is negligible for uncompressed data, but was quite noticeable (~2 seconds) when compressed.

So, all things considered, I decided to not bother with compression right now. We can always add it later.

philpax (approved):
Looks good to me. Would this also enable continuing an existing generation that ran out before its end-of-text?

setzer22:
> Looks good to me. Would this also enable continuing an existing generation that ran out before its end-of-text?

Not directly, but it would be a step in the right direction. Basically, we would need to drop the first tokens in the memory so that we make up space to continue generation.

I'm thinking, adding an API for manipulating the tokens in the memory would lead to very interesting use cases for experimentation 🤔

llama-rs/src/llama.rs
```diff
@@ -575,55 +629,56 @@ impl LlamaModel {
...
-        &self,
-        vocab: &GptVocab,
+    pub fn inference_with_prompt<'a>(
+        &mut self,
+        vocab: &'a GptVocab,
         params: &InferenceParams,
         prompt: &str,
         rng: &mut impl rand::Rng,
+        stop_after_prompt: bool,
+        callback: impl Fn(OutputToken<'a>),
```
mwbryant, line 639: Instead of adding a new parameter to this function would it be better to break this out into 2 separate functions.  One to feed the model prompt data and one to run an inference.   A bool here doesn't feel like the best API to me.  The model would then need to track its own last_n_tokens and maybe more but that doesn't seem unreasonable to me.

We might then be able to get away from the callback and just have the inference return the tokens, giving the user more control over when to run inferencing instead of forcing it to run to completion.  Putting the core loop back into user code.

mwbryant (changes requested):
Great new feature,  I had some thoughts about the API but the functionality seems like it will open a lot of possibilities.

llama-rs/src/llama.rs
philpax: What I was considering was exposing more of the raw inference loop, but wrapping it up in APIs that make it straightforward to use, so that people can choose the behaviour they need. Not sure what that API would look like yet (would need to look at it after work), but I would expect it to enable prompt caching and continued inference.

mwbryant: That sounds great!  I'm coming from the Bevy world so a good API in my opinion gives short running functions that does the minimal unit of work and then gives control back to the caller.  The callback system is inverted to how I would want to use this in projects.

philpax: Yeah, I added the callback as the smallest-change-possible solution to the existing mechanism. I don't want to expose too much of the details around actually driving the model, but I would like to enable users to dig around.

I think the easiest way to experiment with this is to replicate the loop in user code (I think all of the relevant functions are public) and seeing what we can hide behind the scenes while still allowing for flexibility. Suggestions welcome!

setzer22: How about something like this: A `Model::feed_prompt` method you can use to feed the model a sentence. This would not perform any inference, just the prompt.

Then, another method `Model::next_token` that given the current state of the model memory, returns the next token with no external input. You can call this as many times as you want on a loop, react to the results, and stop whenever you want.

Finally, we can keep the original `infer_with_prompt` as a trivial wrapper over the two previous methods.

This new structure would help a lot in untangling the current mess that `infer_with_prompt` is, and I don't see many downsides to it.

What do you think?

philpax: Seems reasonable to me! That also solves a problem I've had with llamacord - there's no way to stop generation when it starts. Having manual control over fetching the next token should sort that out.

Would you be able to feed a prompt after calling `next_token`?

setzer22: Yes, feed prompt or next token could be interleaved in any way the user wants. 😄

setzer22:
Alright, this was some major cleanup session! :smile: 

As discussed, I've broken down the old `infer_with_prompt` function into two functions: `feed_prompt` and `infer_next_token`. This helps untangle the mess the original function was. A `Model` now acts like a "poll"-based API. When you want to produce a token, or feed some more prompt, you just call it. It is now up to the caller to stop on "end of text" or do anything else entirely. :smile:

As part of this refactor, I've also had to move some of the things that were previously declared inside `infer_with_prompt` into the `Model` struct.

Finally, I've refactored the `Memory` structs from the previous iteration of this PR into `InferenceSnapshot`, which stores more things than just the memory (for example, the last_n_tokens and the logits vector). This ensures inference will resume at *the exact same point* where it left off when restoring a snapshot.

llama-rs/src/llama.rs
setzer22: >  Are you planning on building it?

The API? Yes :grin: It's up already!

I had no direct plans, but I'd be super interested in seeing ReAct working with LLaMA using this :eyes:

philpax (changes requested):
Implementation looks good! Just going to make a few suggestions :)

llama-cli/src/main.rs
```diff
@@ -91,15 +94,53 @@ fn main() {
...
+        model.feed_prompt(&vocab, &inference_params, &prompt, |t| {
+            print!("{t}");
+            std::io::stdout().flush().unwrap();
+        });
+        println!();
+
+        // Write the memory to the cache file
+        // SAFETY: no other model functions used inside the block
+        unsafe {
+            let memory = model.get_snapshot();
```
philpax, line 128: Is there a way to make this safe? Something like `memory.with_snapshot(|snapshot| {})` so that access is bounded

llama-rs/src/lib.rs
```diff
@@ -95,6 +109,38 @@ pub struct Vocabulary {
     mapping: Vec<Token>,
 }
 
+#[derive(serde::Serialize)]
+/// A serializable snapshot of the inference process. Can be saved to disk
```
philpax, line 113: Full stop at the end, otherwise rustdoc will render this as "Can be saved to disk Useful for prompt caching"

llama-rs/src/lib.rs
```diff
@@ -156,7 +202,7 @@ pub enum LoadProgress<'a> {
 }
 
 #[derive(Error, Debug)]
-pub enum LoadError {
+pub enum LlamaError {
```
philpax, line 205: Just `Error` should be fine I think, the module should be sufficient scoping

llama-rs/src/lib.rs
```diff
@@ -1142,4 +1148,95 @@ impl Model {
...
         res
     }
+
+    /// Obtains a serializable snapshot of the current inference status. This
+    /// can be used to cache the state of the model and store them into a file.
+    ///
+    /// # Safety
+    ///
+    /// This function provides raw access to the underlying memory owned by the
+    /// ggml context. While the provided `LlamaModelMemory` object is alive, no
```
philpax, line 1158: `InferenceSnapshotRef`

llama-rs/src/lib.rs
```diff
@@ -1142,4 +1148,95 @@ impl Model {
...
+    /// Clears the model's state, so inference can be restarted without having
+    /// to load the weights again.
+    pub fn clear_state(&mut self) {
+        self.memory_k.zero_data();
+        self.memory_v.zero_data();
+        self.n_past = 0;
+        self.last_n_tokens = VecDeque::from(vec![0; self.last_n_tokens.len()]);
+    }
+
+    /// Sets the state of the model, from a previously obtained InfrenceSnapshot
```
philpax, line 1187: `InfrenceSnapshot` -> `InferenceSnapshot`

llama-rs/src/lib.rs
```diff
@@ -1142,4 +1148,95 @@ impl Model {
...
+            memory_k,
+            memory_v,
+            last_n_tokens: self.last_n_tokens.clone(),
+            logits: self.last_logits.clone(),
+        }
+    }
+
+    /// Clears the model's state, so inference can be restarted without having
+    /// to load the weights again.
+    pub fn clear_state(&mut self) {
```
philpax, line 1180: Not a huge fan of this - it seems footgun-y (like it'd be easy to forget to clear the state after you're done). I'd suggest decoupling inference from the model itself, so that you have something like

```rust
let mut session: InferenceSession = model.infer();
session.feed_prompt("blah");
```
where all of the state relevant to this inference is in `InferenceSession`, and the model is purely data. (Bonus points if you have it borrow from `Model`.)

llama-rs/src/lib.rs
```diff
@@ -678,182 +735,132 @@ impl Model {
...
+        vocab: &Vocabulary,
+        params: &InferenceParameters,
+        prompt: &str,
+        rng: &mut impl rand::Rng,
+        callback: impl Fn(OutputToken),
+    ) {
+        self.feed_prompt(vocab, params, prompt, |tk| callback(tk));
 
-            if eot {
+        while self.n_past < self.hparams.n_ctx as usize {
```
philpax, line 794: This loop seems a little weird to me. Maybe have a comment explaining what it's doing for people who look at it to figure out how to structure their own loops?

llama-cli/src/main.rs
setzer22: Basically, this "unsafe" is here out of caution. This function should not be marked unsafe, because the returned object has the correct lifetime and prevents mutating the model for as long as it lives.

The problem here is that tensors inside `ggml` can be mutated through a shared ref `&`. When writing the function to return the slice I realized all this and thought: "Let's just mark this as unsafe for the time being until we figure a better way".

The whole "safety-ness" of the library needs to be reviewed, but I'd do that in another PR. For now, let's just push the requirements to the caller on this one.

llama-rs/src/lib.rs
```diff
@@ -1142,4 +1148,95 @@ impl Model {
...
+        self.n_past = snapshot.npast;
+        self.last_n_tokens = snapshot.last_n_tokens;
+        self.last_logits = snapshot.last_logits;
+
+        Ok(())
+    }
+}
+
+impl<'a> InferenceSnapshotRef<'a> {
+    pub fn write_to_disk(&self, path: impl AsRef<Path>) -> Result<(), LlamaError> {
```
philpax, line 1215: Suggest splitting this into `write(&self, writer: &mut Write)` and `write_to_disk`, where the latter just calls the former with a `File`

llama-rs/src/lib.rs
```diff
@@ -1142,4 +1148,95 @@ impl Model {
...
+            File::create(path).map_err(|err| LlamaError::FailedToWriteMemory(Box::new(err)))?,
+        );
+
+        bincode::serialize_into(writer, &self)
+            .map_err(|err| LlamaError::FailedToReadMemory(Box::new(err)))
+    }
+}
+
+impl InferenceSnapshot {
+    pub fn load_from_disk(path: impl AsRef<Path>) -> Result<Self, LlamaError> {
```
philpax, line 1230: Ditto, but with `Read`

llama-rs/src/lib.rs
```diff
@@ -185,43 +231,50 @@ pub enum LoadError {
     TensorWrongSize { tensor_name: String, path: PathBuf },
     #[error("invalid ftype {ftype} in {path:?}")]
     InvalidFtype { ftype: i32, path: PathBuf },
+    #[error("could not write model memory")]
+    FailedToWriteMemory(Box<dyn std::error::Error>),
+    #[error("could not read model memory")]
+    FailedToReadMemory(Box<dyn std::error::Error>),
```
philpax, line 237: Could we avoid erasing the error type here by adding more cases to the variant and then using `Box<LlamaError>` instead?

llama-cli/src/main.rs
philpax: Fair enough, let's revisit later!

llama-rs/src/lib.rs
setzer22: Hmmmm, you're right :thinking: This requires moving the memory_k and memory_v tensors into another struct though. They should probably be owned by an entirely separate ggml context stored inside the `InferenceSession` to make things easier.

llama-rs/src/lib.rs
setzer22: The idea is to continue inferring tokens until the model returns an EndOfText, or we run out of space. Whatever happens first.

philpax (approved):

xingchensong:
great job!

PR #28, opened 2023-03-16
Remove bindgen

### LINKED ISSUE #26: Consider removing the `bindgen` dependency
#22 and #24 show that `clang` is required as a build dependency because `bindgen` uses it to read in `ggml.h` and produce bindings. We should consider just writing the C bindings ourselves - it shouldn't be too difficult as long as we account for the idiosyncrasies (like #6)

philpax:
Fixes #26.

I removed all references to `ggml_raw` outside of `ggml`, then took the bindgen output, deleted everything that wasn't used, and cleaned it up a little. As you can see, it's not too big.

At the same time, it doesn't look like the original bindgen output is dependent on platform, either. I'm not sure if we should switch over to calling bindgen and committing the output, or maintaining our own C bindings.

Part of me does like the simplicity of knowing *exactly* what we depend on, though.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,28 +12,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "bindgen"
-version = "0.62.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c6720a8b7b2d39dd533285ed438d458f65b31b5c257e6ac7bb3d7e82844dd722"
-dependencies = [
- "bitflags",
- "cexpr",
- "clang-sys",
- "lazy_static",
- "lazycell",
- "log",
- "peeking_take_while",
- "proc-macro2",
- "quote",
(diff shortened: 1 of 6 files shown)


setzer22:
> At the same time, it doesn't look like the original bindgen output is dependent on platform, either. I'm not sure if we should switch over to calling bindgen and committing the output, or maintaining our own C bindings.

I wouldn't be too worried about that. Maybe we can start like this (with the manually maintained bindings), and see where things go from there.

If ggml turns out to be faster than Rust tensor library alternatives and we have to keep it for a long time, then maybe we should find a better solution. But for the time being it sounds like this is the path of least resistance, since it removes a lot of build dependencies (clang, LLVM).

setzer22 (approved):
LGTM, I think we can merge :+1:

philpax:
Alrighty, hit it!

PR #29, opened 2023-03-17
Add basic alpaca REPL mode

darthdeus:
A working REPL prototype with the alpaca prompt. Not sure if we want the alpaca there :thinking:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -115,10 +115,48 @@ dependencies = [
 ]
 
 [[package]]
+name = "clipboard-win"
+version = "4.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7191c27c2357d9b7ef96baac1773290d4ca63b24205b82a3fd8a0637afcf0362"
+dependencies = [
+ "error-code",
+ "str-buf",
+ "winapi",
+]
+
+[[package]]
+name = "dirs-next"
+version = "2.0.0"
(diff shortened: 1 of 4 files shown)


setzer22:
Thanks a lot! :smile: 

I took the liberty to cleanup things a bit since we've been merging some breaking changes.

I think having the alpaca prompt is great, but we should make this configurable, so I changed a bit how the command works:

Instead of ignoring the prompt, we take any prompt string using the usual args (-p, -f). And replace the string `$PROMPT` inside the given prompt, with whatever is read from the line.

Then, I added an `examples/alpaca_prompt.txt` with the following:
```
Below is an instruction that describes a task. Write a response that appropriately completes the request.

### Instruction:

$PROMPT

### Response:
```

I also made it so that the original prompt isn't spit back to you, and instead a little spinner is shown while the prompt is being parsed.

Here's a demo
alpaca_repl_screencap

philpax (approved):
LGTM

RCasatta:
Sorry for the noob question, empty prompt (hitting return at `>>`) returns different casual answers, so is there randomness in the process?

philpax:
Yup - at each step of the LLM, it produces a list of probabilities, and it will sample from that list to produce different outputs. If you'd like to fix the results, you can use a specific `--seed` (it will always deterministically sample according to that seed (with the caveat that it's only for your machine)) or `--top-k 1` (which will always pick the most probable word).

PR #32, opened 2023-03-17
Update GGML

setzer22:
For now, I'm updating GGML by copying the source directly from llama.cpp's repo. We can think of submodule vendoring if these kinds of copies become too frequent, but I'm not super fond of the git submodule UX :sweat_smile: 

I also added a file crediting the original authors, and took the chance to use the file to annotate the specific commit where we took the files from.


philpax (approved):
LGTM

PR #34, opened 2023-03-18
Miscellaneous tweaks

philpax:
Figured it was easier to do all of these in a batch than in individual PRs. Can break them out if required.


setzer22:
LGTM :smile:  These are all things we already discussed so I'll just go ahead and merge :+1:

Issue #27, opened 2023-03-16
Deterministic generations


philpax:
Given the same seed and prompt, the same text should be generated. This will require us to implement a deterministic PRNG (instead of using thread_rng), and to allow specifying a seed. This should also assist in benchmarking.

setzer22:
I would say, for the time being, determinism *given the same hardware and compiler version* is a good enough goal. Going beyond that and trying to make things deterministic across different kinds of hardware is probably going to negatively affect performance.

philpax:
Aye, agreed with setzer - the primary thing I want is to be able to specify the same parameters on the same machine and get the same results. We can think about offering a "fully deterministic" mode later, but as you've mentioned madness lies that way.

setzer22:
I think this can be closed since the rest of the work to get determinism is out-of-scope for now :+1:

PR #36, opened 2023-03-18
Adds a way to specify the seed for RNG

setzer22:
Implements #27 (at least the part we *can* implement, deterministic floating point math is out of scope for the project).

This allows specifying a `--seed` value in `llama-cli`. If no value is passed, the rng is seeded from system entropy like before.

### DIFF
--- a/llama-cli/src/main.rs
+++ b/llama-cli/src/main.rs
@@ -2,7 +2,7 @@ use std::{convert::Infallible, io::Write};
 
 use cli_args::CLI_ARGS;
 use llama_rs::{InferenceParameters, InferenceSnapshot};
-use rand::thread_rng;
+use rand::{thread_rng, SeedableRng};
 
 mod cli_args;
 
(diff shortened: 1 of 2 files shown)


llama-cli/src/main.rs
```diff
@@ -2,7 +2,7 @@ use std::{convert::Infallible, io::Write};
 
 use cli_args::CLI_ARGS;
 use llama_rs::{InferenceParameters, InferenceSnapshot};
-use rand::thread_rng;
+use rand::{thread_rng, SeedableRng};
```
philpax, line 5: `thread_rng` is no longer used

philpax (approved):
Tested with no issues on my machine. Wow, that was easy!

llama-cli/src/main.rs
setzer22: Looks like CI's borked, it didn't catch this: https://github.com/setzer22/llama-rs/actions/runs/4455272439/jobs/7824987492?pr=36

Clippy emitted the warning, but the test didn't faill

philpax: Looks like we need to add `-Dwarnings` to the clippy command

PR #37, opened 2023-03-18
WIP HTTP server for streaming inferences

jack-michaud:
To use the /stream endpoint in llama http with a curl example, send a POST request with a body containing a PredictionRequest in JSON format. Here's an example curl command that does this: 

```
curl -X POST -H "Content-Type: application/json" -d '{"num_predict": 10, "prompt": "This is a test"}' http://localhost:8080/stream
```

Currently, the stream gets piped to the response all at once. I can't figure out how to flush each token with a `flume` stream.

### DIFF
--- a/llama-http/src/cli_args.rs
+++ b/llama-http/src/cli_args.rs
@@ -1,0 +1,79 @@
+use clap::Parser;
+use once_cell::sync::Lazy;
+
+#[derive(Parser, Debug)]
+#[command(author, version, about, long_about = None)]
+pub struct Args {
+    /// The port to listen on
+    #[arg(long, short = 'P', default_value_t = 8080)]
+    pub port: u16,
+
+    /// Where to load the model path from
+    #[arg(long, short = 'm')]
+    pub model_path: String,
+
+    /// The prompt to feed the generator. Prefixes all prompts served.
+    #[arg(long, short = 'p', default_value = None)]
+    pub prompt: Option<String>,
+
+    /// A file to read the prompt from. Takes precedence over `prompt` if set. Prefixes all prompts
+    /// served.
+    #[arg(long, short = 'f', default_value = None)]
+    pub prompt_file: Option<String>,
+
+    /// Sets the number of threads to use
+    #[arg(long, short = 't', default_value_t = num_cpus::get_physical())]
+    pub num_threads: usize,
+
+    /// Sets how many tokens to predict
+    #[arg(long, short = 'n')]
+    pub num_predict: Option<usize>,
+
+    /// Sets the size of the context (in tokens). Allows feeding longer prompts.
+    /// Note that this affects memory. TODO: Unsure how large the limit is.
+    #[arg(long, default_value_t = 512)]
+    pub num_ctx_tokens: usize,
+
+    /// How many tokens from the prompt at a time to feed the network. Does not
+    /// affect generation.
+    #[arg(long, default_value_t = 8)]
+    pub batch_size: usize,
+
+    /// Size of the 'last N' buffer that is used for the `repeat_penalty`
+    /// option. In tokens.
+    #[arg(long, default_value_t = 64)]
+    pub repeat_last_n: usize,
+
+    /// The penalty for repeating tokens. Higher values make the generation less
+    /// likely to get into a loop, but may harm results when repetitive outputs
+    /// are desired.
+    #[arg(long, default_value_t = 1.30)]
+    pub repeat_penalty: f32,
+
+    /// Temperature
+    #[arg(long, default_value_t = 0.80)]
+    pub temp: f32,
+
+    /// Top-K: The top K words by score are kept during sampling.
+    #[arg(long, default_value_t = 40)]
+    pub top_k: usize,
+
+    /// Top-p: The cummulative probability after which no more words are kept
+    /// for sampling.
+    #[arg(long, default_value_t = 0.95)]
+    pub top_p: f32,
+
+    /// Stores a cached prompt at the given path. The same prompt can then be
+    /// loaded from disk using --restore-prompt
+    #[arg(long, default_value = None)]
+    pub cache_prompt: Option<String>,
+
+    /// Restores a cached prompt at the given path, previously using
+    /// --cache-prompt
+    #[arg(long, default_value = None)]
+    pub restore_prompt: Option<String>,
+}
+
+/// CLI args are stored in a lazy static variable so they're accessible from
+/// everywhere. Arguments are parsed on first access.
+pub static CLI_ARGS: Lazy<Args> = Lazy::new(Args::parse);
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,6 +12,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "autocfg"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
+
+[[package]]
 name = "bincode"
 version = "1.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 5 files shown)


philpax:
Huh, odd that the entire stream needs to be consumed. Not sure what's going on there - I thought perhaps `wrap_stream` might not do what we want here, but from what I can tell it should. 

For what it's worth, manually iterating over the stream works for me in `llamacord`: https://github.com/philpax/llamacord/blob/bd63d6b6f689790d03b04672d5a6e87808aba9f7/src/main.rs#L401-L415 (it updates as new tokens come in)

jack-michaud:
> Huh, odd that the entire stream needs to be consumed. Not sure what's going on there - I thought perhaps `wrap_stream` might not do what we want here, but from what I can tell it should. 
> 
> 
> 
> For what it's worth, manually iterating over the stream works for me in `llamacord` (it updates as new tokens come in): 
> 
> 
> 
> https://github.com/philpax/llamacord/blob/bd63d6b6f689790d03b04672d5a6e87808aba9f7/src/main.rs#L401-L415

Funny! It must be the wrap_stream method. It worked with an async mcsp receiver, so maybe I can make a shim with that.

hlhr202:
> Huh, odd that the entire stream needs to be consumed. Not sure what's going on there - I thought perhaps `wrap_stream` might not do what we want here, but from what I can tell it should.
> 
> For what it's worth, manually iterating over the stream works for me in `llamacord` (it updates as new tokens come in):
> 
> https://github.com/philpax/llamacord/blob/bd63d6b6f689790d03b04672d5a6e87808aba9f7/src/main.rs#L401-L415

I m using flume with actix async executor (if not wrong should be tokio behind it)
I follow your approach by spawning a new actix task for looping flume response, but the token is not arrived in real time.
still cant make it work.

image
can help me figure out the problem plz? thx!

jack-michaud:
@hlhr202 is your first actix executor putting tokens onto the stream? 

It would help if I wasn't blocking the thread by synchronously waiting for the model generation to stop :laughing: Async is tough.

hlhr202:
> @hlhr202 is your first actix executor putting tokens onto the stream?
> 
> It would help if I wasn't blocking the thread by synchronously waiting for the model generation to stop 😆 Async is tough.

yes, I create llama instance in my first actix executor, and I give it a receiver so i can pass in inference query from another thread. i also give llama instance a sender so it can pop inference msg to other threads. but still i cant receive the inference msg in real time...
it looks like llama-rs will only put the string in buffer when inferencing, but not throwing it out. thats why we have to flush stdout to see a real time console generation.

jack-michaud:
That sounds similar to the discord bot's architecture. Maybe it's the executor that we are using. I tried creating a session in a tokio executor, and it still wasn't sending threads. The discord bot uses std::thread::spawn.

jack-michaud:
Oh my god. I added a newline to the tokens, and it started sending over the network. The lack of newline is why you need `std::io::stdout()::flush()` in the first place too. All my time in python land is like brain rot for my low level programming

llama-http/src/cli_args.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    /// Top-p: The cummulative probability after which no more words are kept
+    /// for sampling.
+    #[arg(long, default_value_t = 0.95)]
+    pub top_p: f32,
+
+    /// Stores a cached prompt at the given path. The same prompt can then be
+    /// loaded from disk using --restore-prompt
+    #[arg(long, default_value = None)]
+    pub cache_prompt: Option<String>,
```
setzer22, line 69: The cache_prompt argument also seems to be ignored. It doesn't make a lot of sense, at least not as a CLI argument, since the point of the server is to handle multiple requests.

llama-http/src/inference.rs
```diff
@@ -0,0 +1,105 @@
...
+    std::thread::spawn(move || {
+        let args = &*CLI_ARGS;
+
+        // TODO Preload prompt
+
+        // Load model
+        let (mut model, vocabulary) =
+            llama_rs::Model::load(&args.model_path, args.num_ctx_tokens as i32, |_progress| {
+                println!("Loading model...");
+            })
```
setzer22, line 36: Would be good if you use the same code as in main.rs, which shows an indication of progress:

```
[2023-03-20T19:05:01Z INFO  llama_cli] Loaded tensor 8/291
[2023-03-20T19:05:01Z INFO  llama_cli] Loaded tensor 16/291
[2023-03-20T19:05:02Z INFO  llama_cli] Loaded tensor 24/291
[2023-03-20T19:05:03Z INFO  llama_cli] Loaded tensor 32/291
```

Using the `log` crate for the server sounds like it would also be a good idea. It's even more important for a server to have timestamped logs and offer configurable output levels.

setzer22 (commented):
LGTM :smile:  Other than few very minor comments. But I don't have that much experience in creating Rust server backends, so I'll let others with more experience comment on that.

I think having an HTTP server like this really important to allow new use cases for `llama-rs`, thanks a lot for working on this!

setzer22:
> Namely the model context does not seem to be reset between requests

The best way to handle this is to create one new `InferenceSession` per request. An inferenc session entirely encapsulates the mutable state of the transformer, so managing these is the key to keep per-user state. You can create a new session every time and it will "forget" about all previous input, or even offer an API where users can create sessions and submit multiple requests to it.

And we will probably need some sort of Mutex thing to ensure not too many requests are started in parallel. Maybe our http framework already supports this? For most machines, the limit would be attending 1 inference request at a time.

Also keep in mind each inference session adds ~500MB to memory consumption (it stores the full context window tensors), so we need to make sure to clean them up and to not start too many of them.

llama-http/src/cli_args.rs
```diff
@@ -0,0 +1,79 @@
...
+    pub model_path: String,
+
+    /// The prompt to feed the generator. Prefixes all prompts served.
+    #[arg(long, short = 'p', default_value = None)]
+    pub prompt: Option<String>,
+
+    /// A file to read the prompt from. Takes precedence over `prompt` if set. Prefixes all prompts
+    /// served.
+    #[arg(long, short = 'f', default_value = None)]
+    pub prompt_file: Option<String>,
```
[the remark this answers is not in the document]
jack-michaud: Great call, thank you 5906227

llama-http/src/cli_args.rs
jack-michaud: Removed in 5906227

llama-http/src/inference.rs
jack-michaud: Agreed, added in 541b498 and 7b52b83

jack-michaud:
I've incorporated your feedback, @setzer22. I also now have a separate inference session for each request.


I've created a an `InferenceSessionManager` whose purpose is to limit the amount of sessions created (i.e. lock until one is available). This is created in the "Inference Thread" (done), then an inference session for each "Inference Request Thread" will be created from that manager. Currently, there is no separate inference request thread that handles each request; I haven't gotten to that point yet since ggml tensors cannot be sent between threads safely (wrapping with a Arc + Mutex didn't seem to help, but I wouldn't be surprised if there was user error here). 

image

(This is my imagined architecture: again, "Inference Request Thread" does not exist yet.)

I'd suggest merging in the single threaded state (since most machines will only handle one session at a time) and then add this in a later PR along with a `max_requests` to CLI_ARGS:

```rs
/// Maximum concurrent requests.
/// Keep in mind that each concurrent request will have
/// its own model instance, so this will affect memory usage.
#[arg(long, default_value_t = 1)]
pub max_requests: usize,
```

KerfuffleV2:
Just so it doesn't come out of the blue, I've been looking at doing something related. I've been considering creating a backend that would be transport-agnostic (the idea is just to use Tokio channels or something and then other stuff could potentially use it for implementing HTTP or whatever).

My idea is just to have a task that keeps track of jobs or sessions for the model and can interpret them in a round robin kind of way (initially at first), which would allow having multiple inference sessions/models going at the same time.

The big thing I want to do is have a way to easily pause/snapsnot sessions, specifically when it runs into certain tokens. This should allow stuff like setting bookmarks and rewinding which I think one can do some pretty interesting things with...

This is a sketch of the kind of interface I was thinking of: https://github.com/KerfuffleV2/llm-backend/blob/main/src/types.rs

I'm pretty sure it can be done as an independent project. Anyway, hope this doesn't seem like it's trying to steal your thunder or anything like that. (Also, if there's anything you can find a use for in that repo, feel free.)

philpax:
Hiya! I'm going to close this PR as there's been some pretty major changes in the meantime. However, I think there's still very much a place for an open-source HTTP inference server built on top of `llm`.

If you're still interested in working on this, I'd love to see this reimplemented as a separate repository that targets `llm`.

PR #39, opened 2023-03-18
Implement latest llama.cpp tokenizer

philpax:
Implements the latest tokenizer from https://github.com/ggerganov/llama.cpp/commit/c9f670a17755311aa28c411f5c7f3c8c05434770 and https://github.com/ggerganov/llama.cpp/commit/a81d0c2a171a4446e6a21a3ec74a0c0768d71184

### DIFF
--- a/llama-rs/src/lib.rs
+++ b/llama-rs/src/lib.rs
@@ -116,6 +116,9 @@ pub struct Vocabulary {
 
     /// Maps a token to a token id
     token_to_id: HashMap<Token, TokenId>,
+
+    /// The longest token in this vocabulary
+    max_token_length: usize,
 }
 
 #[derive(serde::Serialize)]
@@ -349,10 +352,12 @@ impl Model {
         let vocab = {
             let mut id_to_token = vec![];
             let mut token_to_id = HashMap::new();
+            let mut max_token_length = 0;
 
             for i in 0..hparams.n_vocab {
                 let len = read_i32(&mut reader)?;
                 if let Ok(word) = read_string(&mut reader, len as usize) {
+                    max_token_length = max_token_length.max(word.len());
                     id_to_token.push(word.clone());
                     token_to_id.insert(word, i);
                 } else {
@@ -366,6 +371,7 @@ impl Model {
             Vocabulary {
                 id_to_token,
                 token_to_id,
+                max_token_length,
             }
         };
 
@@ -1116,16 +1122,13 @@ impl Model {
         text: &str,
         bos: bool,
     ) -> Result<Vec<TokenId>, InferenceError> {
-        // TODO: Calculate this constant from the vocabulary
-        const MAX_TOKEN_LENGTH: usize = 18;
-
         let len = text.len();
 
         let mut score = vec![0usize; len + 1];
         let mut prev = vec![TokenId::default(); len + 1];
 
         for i in 0..len {
-            let max_len = (len - i).min(MAX_TOKEN_LENGTH);
+            let max_len = (len - i).min(vocab.max_token_length);
             for sub_len in 1..=max_len {
                 let sub = &text.as_bytes()[i..i + sub_len];
                 let Ok(sub) = std::str::from_utf8(sub) else { continue; };
--- a/llama-rs/src/lib.rs
+++ b/llama-rs/src/lib.rs
@@ -110,10 +110,12 @@ impl Default for InferenceParameters {
 type TokenId = i32;
 type Token = String;
 
-#[derive(Default)]
 pub struct Vocabulary {
-    /// Maps every integer (index) token id to its corresponding string
-    mapping: Vec<Token>,
+    /// Maps every integer (index) token id to its corresponding token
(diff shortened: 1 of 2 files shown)


setzer22 (approved):
Looks good. I have tried a few examples with a fixed seed between this and the main branch with surprising results. For English and Spanish, the results are always exactly identical, but for French I've observed difference between the new and old tokenizer.

For example, given the command: `cargo run --release -- -m /data/Llama/LLaMA/7B/ggml-model-q4_0.bin --num-ctx-tokens 2048 -p "Je ne veux pas travailler, je ne veux pas déjeuner" --seed 1235`

This result with the new tokenizer (this branch):
```
Je ne veux pas travailler, je ne veux pas déjeuner. Je suis encore étudiant.
I do not want to work (or rather: «je n'ai plus envie de travailler»), I don’t feel like studying either; the time has come for me to get out of here![end of text]
```

And this is the result I get `main` (quite crazy :rofl:):
```
Je ne veux pas travailler, je ne veux pas déjeuner
It's hard to believe that it is almost time for the Christmas holidays. And with them come all those annual predictions about retail sales during this season... This year will be a "very good" one; or at least as good if not better than last years which was also very good (in fact I hear there were some record-breaking weeks). Anyway you look at things people are still going out and buying stuff in droves....which means we should see another solid quarter of growth when they report again next month - despite any gloomy forecasts by pundits! The stock market agrees too..it appears that many analysts have been revising their earnings estimates upwards lately rather then down so far.
And speaking of strong earning reports...Apple(AAPL) has just reported yet ANOTHER spectacularly profitable quarter! It really does seem like everyone else must finally admit what Steve Jobs told us earlier: That Apple IS running on iCloud now :) Now if only someone could give Mr JBP more room/space around him he might actually BE able to run anything :-) Still no matter how great these results always are (or whatever happens with its shares price wise), Wall Street continues to treat AAPL as an underperformer relative to other tech darlings due mainly to fear over hardware supply chain constraints etc..... So let me ask anyone who follows such companies closely: Who carries MORE cash? Google Inc.(GOOG)? Facebook?(FB?) Microsoft Corp.(MSFT ) ? OR APPL??? If memory serves correct GOOGL and FB both carry close to $30 billion while MSFT seems content to sit pretty near halfway between these two giants carrying northward of 25 billiion$ !!! What amuses me most though here isn't even this but why can't investors understand simple math?? Even if your share count increased say from lets pretend 1bn to slightly below 4bln do ya think it would make much difference either way financially IF YOUR SALES TRILLIONS OF DOLLARS AND PROFIT IN THE BILLONS ALREADY!!! You guys don't need anymore $$$$$. We get it already!. Sure maybe the number crunchers find a few extra millions somewhere everytime management announces guidance increases ;) But once u cross certain thresholds getting incrementally more money becomes irrelevant because unless ur selling tons of products for less than cost (a la Amazon)(AMZN); Or unless your manufacturing costs drop magically through technology advances enabling lower margins without sacrificing quality!(like Motorola Inc.)(MMI) U simply cannot compete anyway since you lack scaled economies enough . Anyways before closing todays post Id wanna point out something funny regarding this latest round of earnings releases:) Aside form reporting healthy gains ALL THREE TECH GIANTS mentioned above seemed to perform similarly well altho Google came second best behind Apples incredible iPhone unit sellthrough rate :(. Go figure. In case yah missed my earlier posting google was expecting iphone users to switch allegiances ;-) Well anyway thats it till after xmas folks Have a happy Thanksgiving & enjoy ur day off tomorrow! See you back here later tonite @ 6pm est[end of text]
```

This is purely anecdota evidence of course, but I thought it'd be good to bring it up. If anything, the fact that using the new tokenizer, the model actually generates a sentence in french gives me a bit more confidence that the prompt was properly encoded. But who knows. Over on the C++ side people have been reporting this new tokenizer works better, so I'd say we can merge this.

philpax:
Sounds good to me, up to you to merge 🙏

llama-rs/src/lib.rs
```diff
@@ -110,10 +110,15 @@ impl Default for InferenceParameters {
 type TokenId = i32;
 type Token = String;
 
-#[derive(Default)]
 pub struct Vocabulary {
-    /// Maps every integer (index) token id to its corresponding string
-    mapping: Vec<Token>,
+    /// Maps every integer (index) token id to its corresponding token
+    id_to_token: Vec<Token>,
```
KerfuffleV2, line 115: If this the ids and same order as the existing parser, it seems like it would be very easy to support both and just allow selecting between parsers with an option. That way there really wouldn't be a cost to merging this.

setzer22: We should actually bring the current tokenizer changes from llama.cpp, because both this and the old version of the tokenizer are probably wrong, so I think there's no need to keep both (and offer a configuration flag) when the user probably just wants the one that is correct.

setzer22:
@philpax This looks good to merge, but you seem to have some setting enabled that doesn't let me add commits to the branch :thinking: Can you take care of the conflicts (it's just a minor thing) and merge?

PR #41, opened 2023-03-18
Session caching CLI

### LINKED ISSUE #38: Replace prompt caching with session caching in the CLI
(Will take this soon)

At present, we have `--cache-prompt` and `--restore-prompt`, but these are a little ambiguous in how they work. The former will exit immediately after saving the prompt, and the latter can actually be used with any prompt (not just the one that was saved to disk).

To better communicate what they do and to make them more general, I propose replacing them with `--load-session`, `--save-session` and `--persist-session` (which is an alias for loading and saving to the same path).

- `--load-session` is identical to `--restore-prompt` in that it loads a saved inference snapshot, but it better communicates what it's doing.
- `--save-session` will save the results of inference to disk, similar to `--cache-prompt`, but it will also include whatever was inferred, allowing you to continue on from a response. `--cache-prompt PATH` is equivalent to `--save-session PATH -n 0`. (This could be documented, or another flag could be added... but it almost feels like another "mode" to me. Should figure out how we want to do that for #29, too.)
- `--persist-session` loads a session from an existing path (if it exists) and saves to the path afterwards.

This would allow you to have ongoing conversations over an extended period of time:
```bash
llama-cli --persist-session conversation.llama -p "How do I make bread?"
...
llama-cli --persist-session conversation.llama -p "How long should I let the dough rest at room temperature?"
...
llama-cli --persist-session conversation.llama -p "Can I keep the dough in the fridge?"
```

philpax:
Fixes #38.

I've had to make a few slightly controversial changes here:

- `InferenceSession` now stores all tokens that have been processed, not just the last N tokens
- `inference_with_prompt` now plays back all tokens that have been processed
- I've simplified the loop in `inference_with_prompt` (I realised that the `while` condition was unnecessary since we can just return the error anyway)

On the plus side, it basically works as you'd expect:

```
> cargo run --release -- --model-path ggml-alpaca-7b-q4.bin -t 8 -p "This is why I love open-source software: " -n 16 --persist-session test.llama
# [...]
This is why I love open-source software: 58 people in one room, all working together to make a project better.
[2023-03-18T16:51:23Z INFO  llama_cli] Successfully wrote session to "test.llama"

> cargo run --release -- --model-path ggml-alpaca-7b-q4.bin -t 8 -p " Those 58 people can change the world, together, and here's how: " -n 16 --persist-session test.llama
# [...]
[2023-03-18T16:52:01Z INFO  llama_cli] Loaded inference session from "test.llama"
This is why I love open-source software: 58 people in one room, all working together to make a project better. Those 58 people can change the world, together, and here's how: 1) they are not doing it for money (they could be);
[2023-03-18T16:52:07Z INFO  llama_cli] Successfully wrote session to "test.llama"
```


llama-rs/src/lib.rs
```diff
@@ -1211,21 +1241,21 @@ impl InferenceSession {
...
-            Ok(OutputToken::Token(&vocab.mapping[next_token as usize]))
-        }
+        Ok(OutputToken::from_id(vocab, next_token))
     }
 
     // todo: see if we can reduce the arguments here somehow - consolidate model and vocab maybe?
+    /// Helper function to run inference with this session and the given model and vocabulary.
+    ///
+    /// Note that this will "play back" all existing tokens in the session. If this is not desired
+    /// behaviour, consider implementing your own inference loop to customize the behavior.
```
setzer22, line 1258: Given that the contents of `inference_with_prompt` are not exactly trivial, I'd say this behavior should be controlled by a function parameter.

The most interesting use case for prompt caching is something like Bing, where they feed the model a giant prompt to the model (this one is cached, so you save a lot of processing time), plus the user query at the end. In that case, you don't want to print the contents of the "pre-prompt", just the contents of the prompt you're currently feeding the system. But other than that the `inference_with_prompt` function works perfectly fine, so it would be nice if users that want to do this weren't forced to reimplement their own loops.

setzer22 (commented):
Looks good overall! Just have a couple of comments that I feel are worth discussing

llama-cli/src/cli_args.rs
```diff
@@ -53,28 +54,34 @@ pub struct Args {
...
     /// Specifies the seed to use during sampling. Note that, depending on
     /// hardware, the same seed may lead to different results on two separate
     /// machines.
     #[arg(long, default_value = None)]
     pub seed: Option<u64>,
 }
-
-/// CLI args are stored in a lazy static variable so they're accessible from
-/// everywhere. Arguments are parsed on first access.
-pub static CLI_ARGS: Lazy<Args> = Lazy::new(Args::parse);
```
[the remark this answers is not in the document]
philpax: Mostly just that: it wasn't used, and the application is currently a single file. If you want me to restore it, I can.

llama-rs/src/lib.rs
philpax: Mm, yeah, I'm not sure what the best course of action here is. `inference_with_prompt` is getting a little unwieldy because it's hard to accommodate everyone's use cases. If you think a boolean's the right call here, I'll add it.

setzer22: But I agree, let's make sure we don't add too much bloat and direct people to the lower level functions. But just for this, since the function used to work the other way, let's add the boolean just to make transition easier. I generally don't like breaking changes. Even though this code doesn't have lots of users, it might be annoying if we have to tell people they need to drop down to the lower level API now when they could use this just fine before.

philpax: Fair enough, will do in a bit!

KerfuffleV2:
I've been messing around with adding compression to this. Weirdly, the natural approach of just wrapping the reader/write in compression functions is unbearably slow. However, compressing to a buffer directly is basically instant.

Something pretty weird is going on.

KerfuffleV2:
@philpax 

Pullception: https://github.com/philpax/llama-rs/pull/1

Commenting here in case you want to keep discussion in one place.

setzer22 (approved):
Looks good :+1: I think this one's ready to merge now

PR #47, opened 2023-03-19
feat: record inference stats

### LINKED ISSUE #46: Reporting model stats
Hey, in llama.cpp, there are some useful output for telling the model predict / evaluation stats:

https://github.com/ggerganov/llama.cpp/blob/da5303c1ea68aa19db829c634f1e10d08d409680/main.cpp#L1086-L1095

Can we also export such information after the model run?

I have a local change which produces output like this:

```
[2023-03-19T21:59:46Z INFO  llama_cli] Model size = 4017.27 MB / num tensors = 291
[2023-03-19T21:59:46Z INFO  llama_cli] Model fully loaded!
<prompt> <predict output>
feed_prompt_duration: 1533ms, prompt_tokens: 10, predict_duration: 22656ms, predict_tokens: 86, per_token_duration: 263.442ms
```

bcho:
This pull request added a simple struct `InferenceStats` for capturing the model evaluation stats. This struct can generate a similar output like llama.cpp . Closes #46 

Example output:

```
feed_prompt_duration: 1615ms
prompt_tokens: 10
predict_duration: 6917ms
predict_tokens: 30
per_token_duration: 230.567ms
```

### DIFF
--- a/llama-rs/src/lib.rs
+++ b/llama-rs/src/lib.rs
@@ -5,6 +5,7 @@ use std::{
     fmt::Display,
     io::{BufRead, Read, Seek, SeekFrom},
     path::{Path, PathBuf},
+    time,
 };
 
 use thiserror::Error;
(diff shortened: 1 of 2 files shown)


llama-rs/src/lib.rs
```diff
@@ -107,6 +108,38 @@ impl Default for InferenceParameters {
...
+            predict_tokens: 0,
+        }
+    }
+}
+
+impl Display for InferenceStats {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(
+            f,
+            "feed_prompt_duration: {}ms, prompt_tokens: {}, predict_duration: {}ms, predict_tokens: {}, per_token_duration: {:.3}ms",
```
setzer22, line 133: Minor nit: Wouldn't it make more sense to print these in separate lines?

What do others think?

setzer22 (approved):
Thanks a lot for sharing! :smile: It was always bugging me that I couldn't see the times per token with this.

At first I thought maybe we needed to integrate this into the low-level functions `feed_prompt` and `predict_next_token`, but on second thought I prefer this version because it's simple, and if users are building a lower level loop using the other functions, they can do their own time measurements too, so there's no point in overengineering things.

Just have a very minor nitpick about newlines :thinking: but overall LGTM

llama-rs/src/lib.rs
bcho: np, updated, PTAL

setzer22 (approved):
Thanks again! We can merge once CI passes :)

PR #50, opened 2023-03-21
chore: update readme for easier running of models

hhamud:
Updated the README so that users don't have to constantly use cargo run -- every time they use the binary.


hhamud:
> Hm, I think the preferable solution for this is to `cargo install` it so that it's available in your PATH?
> 
> We're planning on getting this into `crates.io`, so all you'll need to do is `cargo install llama-cli` and then you'll be able to do `llama-cli ...`.

Good point, how about changing it to cargo install --path for now?

Issue #51, opened 2023-03-21
Use F16 for memory_k and memory_v


univerz:
see https://github.com/ggerganov/llama.cpp/pull/154 & https://github.com/ggerganov/llama.cpp/pull/270#issuecomment-1477175472

KerfuffleV2:
This is completed. You can now use `--float16`.

PR #54, opened 2023-03-21
Add --float16 (was --float-16) to llama-cli and associated support to llama-rs

KerfuffleV2:
This change adds a `--float-16` flag which allows using a 16 bit float for the model memory key and value tensors. 

The `--float-16` flag enables it for both K and V, however internally it is possible to set it independently (I don't know why you'd want to, but...)

In addition to saving memory, this change only reduces the size of the cached prompt by about 50% when enabled.

Should complete #51.

### DIFF
--- a/llama-cli/src/cli_args.rs
+++ b/llama-cli/src/cli_args.rs
@@ -81,7 +81,7 @@ pub struct Args {
     /// Use 16-bit floats for model memory key and value. Ignored when restoring
     /// from the cache.
     #[arg(long, default_value_t = false)]
-    pub float_16: bool,
+    pub float16: bool,
 }
 
 /// CLI args are stored in a lazy static variable so they're accessible from
--- a/llama-cli/src/cli_args.rs
+++ b/llama-cli/src/cli_args.rs
@@ -77,6 +77,11 @@ pub struct Args {
     /// machines.
     #[arg(long, default_value = None)]
     pub seed: Option<u64>,
+
+    /// Use 16-bit floats for model memory key and value. Ignored when restoring
+    /// from the cache.
+    #[arg(long, default_value_t = false)]
+    pub float_16: bool,
 }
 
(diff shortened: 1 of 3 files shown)


RCasatta:
I am not sure if I am understanding/testing this right but the following command:

```
cargo run --release -- --model-path ~/Downloads/LLaMA/7B/ggml-model-f16.bin -f examples/alpaca_prompt.txt -R --float-16
```

use the same amount of memory (as reported by htop), ~12G on my machine, with or without `--float-16`

philpax (approved):
LGTM if it works. Only things I'd suggest:
- use `Serialize` and `Deserialize` instead of `serde::Serialize`/`serde::Deserialize`
- use `--float16` instead of `--float-16`

KerfuffleV2:
@RCasatta 

The default context size is 512, which means with 32bit it would be around 512MB, with 16bit it would be 256MB. That may not be a big enough change to notice from htop depending on how it's rounding the memory sizes. Try with `--num-ctx-tokens 2048` and you should notice a bigger difference. You can also try using `--cache-prompt FILENAME` with a prompt specified and observe how the output file changes size depending on whether the flag is on or off.

***

@philpax 

No problem. I can make those changes, however other code (the caching stuff) is using `serde::Deserialize` and `serde::Serialize`. So I would need to change the imports to bring it into scope and then it probably would make sense to change the other code using Serde to use that as well. I was trying to keep the scope of the changes small.

Changing the name of the flag is certainly no issue. What do you want me to do about the Serde stuff?

setzer22:
> What do you want me to do about the Serde stuff?

The code is using the same style of the other structs. So I'd say we can just merge this and sneak in a change to `use serde::{Serialize, Deserialize}` in another PR.

I'll just have a quick look at the diff first :eyes:

setzer22:
Thanks a lot for this BTW! :smile:

KerfuffleV2:
Happy to contribute!

PR #61, opened 2023-03-23
Support latest (breaking) changes to GGML file format

RoyVorster:
Latest ggml models have:
- Different magic
- A format version
- Scores for tokens (currently parsed but unused)

This PR maintains support for older ('legacy') models
A model downloaded and converted on latest master in `llama.cpp` now works with `llama-rs` but the output seems qualitatively worse. Haven't played around much with `llama-rs` though so it's hard to say whether this is a regression.

### DIFF
--- a/llama-rs/src/lib.rs
+++ b/llama-rs/src/lib.rs
@@ -385,15 +385,15 @@ impl Model {
         }
 
         fn read_i32(reader: &mut impl BufRead) -> Result<i32, LoadError> {
-            return Ok(i32::from_le_bytes(read_bytes::<4>(reader)?));
+            Ok(i32::from_le_bytes(read_bytes::<4>(reader)?))
         }
 
         fn read_u32(reader: &mut impl BufRead) -> Result<u32, LoadError> {
-            return Ok(u32::from_le_bytes(read_bytes::<4>(reader)?));
+            Ok(u32::from_le_bytes(read_bytes::<4>(reader)?))
         }
 
         fn read_f32(reader: &mut impl BufRead) -> Result<f32, LoadError> {
-            return Ok(f32::from_le_bytes(read_bytes::<4>(reader)?));
+            Ok(f32::from_le_bytes(read_bytes::<4>(reader)?))
         }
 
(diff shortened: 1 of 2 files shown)


llama-rs/src/lib.rs
```diff
@@ -297,6 +311,28 @@ macro_rules! mulf {
     };
 }
 
+trait FromLeBytes {
```
RoyVorster, line 314: I couldn't really figure out a cleaner way to do a trait bound for a bunch of different 32bit long numerical primitives.

setzer22:
Hi! Thanks a lot for the PR :smile: 

> A model downloaded and converted on latest master in llama.cpp now works with llama-rs but the output seems qualitatively worse. 

That's really strange. IIRC the only changes to the output are that now vocab scores are embedded in the model. Weights should be exactly the same.

Can you make a test with `--seed` to ensure results are the same between this and the `main` branch?

llama-rs/src/lib.rs
```diff
@@ -147,6 +147,16 @@ type Token = String;
...
+    scores: Vec<f32>,
+}
+
+impl Vocabulary {
+    fn with_capacity(n_vocab: usize) -> Self {
+        Self {
+            mapping: Vec::<Token>::with_capacity(n_vocab),
+            scores: Vec::<f32>::with_capacity(n_vocab),
+        }
+    }
```
setzer22, line 159: Unfortunately there's now a conflict between this and the recently merged #39. Shouldn't be hard to fix, but let me know if you have questions!

setzer22 (approved):
Looks good! Thanks again for the changes. Can you sort out the merge conflict and the formatting / clippy issues from CI? 

Once that's taken care of, we're ready to merge :smile:

RoyVorster:
Yup, will doublecheck output vs. main branch on an older model to ensure all is okay.

RoyVorster:
Yeah looks good to me. @setzer22 can you approve the workflow?

setzer22:
Merged! Took care of some minor clippy lints myself.

Thanks again :)

PR #64, opened 2023-03-23
Add --bias-tokens and --dump-prompt-tokens flags and associated backend changes.

KerfuffleV2:
`--bias-tokens` allows passing a list of token ids and float point biases. This can be used to emulate llama.cpp's `--ignore-eos` feature in a more flexible way. For example using `--bias-tokens 2=-1` will prevent the EOD token from being generated and the model will just ramble on until it hits the context limit.

`--dump-prompt-tokens` will just dump the tokens for the prompt after loading, in two formats: First just as a list of token ids and then as a list with the associated string the token came from. For example, with prompt `What a kerfuffle!`, you'd get the output:

    === Dumping prompt tokens:
    5618, 263, 13023, 21154, 600, 280, 29991
    "What":5618, " a":263, " ker":13023, "fu":21154, "ff":600, "le":280, "!":29991

***

Some minor backend changes were necessary to make this possible but the behavior of the existing code should remain unchanged. One thing I did do was move the `tokenize` method to `Vocabulary` and changed it to return `Vec<(&str, TokenId)>`. The `tokenize` method in the session remains as a wrapper to that which just strips out the string part.

### DIFF
--- a/llama-rs/src/lib.rs
+++ b/llama-rs/src/lib.rs
@@ -5,6 +5,7 @@ use std::{
     fmt::Display,
     io::{BufRead, Read, Seek, SeekFrom},
     path::{Path, PathBuf},
+    str::FromStr,
     time,
 };
 
@@ -131,6 +132,7 @@ pub struct InferenceParameters {
     pub top_p: f32,
     pub repeat_penalty: f32,
     pub temp: f32,
+    pub bias_tokens: TokenBias,
 }
 
 impl Default for InferenceParameters {
@@ -142,6 +144,7 @@ impl Default for InferenceParameters {
             top_p: 0.95,
             repeat_penalty: 1.30,
             temp: 0.80,
+            bias_tokens: TokenBias::default(),
         }
     }
 }
@@ -243,6 +246,55 @@ impl Display for OutputToken<'_> {
                 OutputToken::EndOfText => "[end of text]",
             }
         )
+    }
+}
+
+#[derive(Default, Clone, Debug, PartialEq)]
+pub struct TokenBias(Vec<(TokenId, f32)>);
+
+impl TokenBias {
+    pub fn new(mut v: Vec<(TokenId, f32)>) -> Self {
+        v.sort_by_cached_key(|(tid, _)| *tid);
+        v.dedup_by_key(|(tid, _)| *tid);
+        Self(v)
+    }
+
+    pub fn get(&self, tid: TokenId) -> f32 {
+        self.0
+            .binary_search_by_key(&tid, |(tid, _)| *tid)
+            .map_or(0.0, |idx| self.0[idx].1)
+    }
+}
+
+impl FromStr for TokenBias {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let x = s
+            .split(',')
+            .map(|kv| {
+                let (k, v) = kv
+                    .trim()
+                    .split_once('=')
+                    .ok_or_else(|| "Missing '=' in bias item".to_owned())?;
+                let tid: TokenId = k
+                    .trim()
+                    .parse()
+                    .map_err(|e: std::num::ParseIntError| e.to_string())?;
+                let bias: f32 = v
+                    .trim()
+                    .parse()
+                    .map_err(|e: std::num::ParseFloatError| e.to_string())?;
+                Result::<_, String>::Ok((tid, bias))
+            })
+            .collect::<Result<_, _>>()?;
+        Ok(TokenBias::new(x))
+    }
+}
+
+impl std::fmt::Display for TokenBias {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self.0)
     }
 }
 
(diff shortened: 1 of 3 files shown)


KerfuffleV2:
It's worth noting that the tokens specified with `--bias-tokens` aren't subject to the repeat penalty calculation. I figured the simplest possible behavior was just to use the absolute value specified.

There are other possible approaches like letting the repeat penalty be applied to the specified bias value. Also, I'm not sure what the correct range for those values would be, but I'd guess it's probably -1 to +1?

KerfuffleV2:
I'll rebase this if it seems like it's something you'd generally be interested in accepting.

setzer22:
> I'll rebase this if it seems like it's something you'd generally be interested in accepting.

Yes please! I need to go over the changes but this feature is quite nice and I was thinking of adding something similar :smile:

llama-rs/src/lib.rs
```diff
@@ -823,18 +875,24 @@ impl Model {
...
             for (i, &logit) in logits.iter().enumerate() {
+                let tid = i as TokenId;
+                let bias = params.bias_tokens.get(tid);
+
                 // repetition penalty from CTRL paper (https://arxiv.org/abs/1909.05858)
                 // credit https://github.com/facebookresearch/llama/compare/main...shawwn:llama:main
-                if session.last_n_tokens.contains(&(i as TokenId)) {
+                let val = if bias != 0.0 {
+                    bias
+                } else if session.last_n_tokens.contains(&tid) {
```
setzer22, line 957: Calling it `bias` sounds a bit weird. Generally a bias is something that is added to the original value. But here, the bias entirely replaces the value?

llama-rs/src/lib.rs
```diff
@@ -823,18 +875,24 @@ impl Model {
         {
             let scale = 1.0 / params.temp;
             for (i, &logit) in logits.iter().enumerate() {
+                let tid = i as TokenId;
+                let bias = params.bias_tokens.get(tid);
```
setzer22, line 951: Would be better if this returns an option instead of 0.0, when the token has no bias.

KerfuffleV2:
@setzer22 Seems like it still works. The tokens are a little different with the new parser (which isn't too surprising):

    === Dumping prompt tokens:
    5618, 263, 13023, 21154, 600, 280, 29991
    "What":5618, " a":263, " ker":13023, "fu":21154, "ff":600, "le":280, "!":29991

setzer22 (commented):
Looks good overall! Just a couple of minor things to discuss :smile:

llama-rs/src/lib.rs
KerfuffleV2: I called it that based on the `logit_bias` field in the TextSynth API: https://textsynth.com/documentation.html#completions

You can call it whatever you want (except late for dinner!) What would you prefer the name to be?

llama-rs/src/lib.rs
KerfuffleV2: Changed. I also change the binding there to be called `logit_override`. I don't know if you want to change the name of the CLI option or struct fields, if so just let me know what you want instead and I can easily make the change.

setzer22: Hmm tough call :thinking: I think the term "token bias" would be a bit more relatable to users. But logit override is more accurate. I think we can do `--logit-overrides` :+1: If you feel that's a good name too. 

Also, thinking about it, would you mind adding a `--ignore-eos` option that simply serves as an alias for the new flag to set the network to ignore the eos? Someone coming from llama.cpp might not realize we have the more general option so it's good to also have the other one.

setzer22: Ok, yeah, let's not overthink this. Your plan sounds good :slightly_smiling_face:

KerfuffleV2: All set. Interestingly, setting the bias to -1 seems to prevent a token from ever being selected, but setting it to 1 (or 2) doesn't seem to have a dramatic effect. I set it to 100 and it immediately generated the token though.

I can confirm disabling EOS works, although the output can become increasingly less relevant...

***

Here we have a never ending fairy tale about a fluffy fox and his adventures. Follow him on the road to find himself, meet some amazing characters along the way like King Kai or Jax! He will always be your companion until you complete all of these levels full with challenges that would test even the greatest strategist's skillfulness in this great strategy game by G-2 Games Inc..
You start as a fox cub who has big dreams but no where for them go yet... So it is time to travel the open world map and explore lots and lots different locations from desert cannibals to mountainous areas where giant snow cats rule over the area!. You get special powerups throughout the journey so make sure not miss those while exploring each one of them! Collect golden apples which grant extra points when collected into energy potions (think Mario level up system) and use the coins wisely because they give discount at shops located around town (you buy items using the same virtual currency). Fight battles against wolves guardians or kobolds. Unlock secret rooms and more !!!
The main objective here being finding yourself among other things suchlike solving puzzle riddels through out the quest to solve bigger mystery storyline later.... The Fox character itself looks really unique too - imagine an Aryan Fox with fluffiness everywhere just makes my day ;) . The controls are intuitive too ,so once I got used to how he moves around maps etc., everything became fluid again,. Everything comes together brilliantly and made me want more !! It also features nice comic graphics/characters and soundtrack so you could say there was nothing missing from this game except perhaps for multiplayer feature :P..... But overall i feel this is truly worth recommending if any fan of games likes good looking cartoony artwork ,strategizing mind challenge type of casual RPG'S then definitely go try Flufflefox & His Friends today !! :) .. Hope u enjoy playing ! Cheers everyone .... See ya next year ... Ummm What?! Just KIDDING DON'T WORRY!!! LOLZ Good luck Adventuring till now XDD . PLAY FOR FUN NOT MONEY AND LEARN LOVE EVERTHING IN THE UNIverse ASAPLY TIME IS RUNNING OUT CURRENT YEAR 48357 DAYS LEFT TO GO BEFORE IT END ONCE SO GET PLANETARIIUM NOWWW *happy dance* before its gone xoxoxo PS: For anyone reading THIS REVIEWS DOESNT SEEMS LIKE IT CONTAIN SPOILERS BUT IF BY CHANCE ANY1 STILL CARES /CANT SAY SHUOD HAHV AVREEZ ARBSTRAKTIONALITY VIKS XxxxxxxxxxLMALLX OV VEEERRRRLYYYY LAATTTTEEEXXXXXX"
Great fun to play. Highly addictable. Great music track though somewhat short lived experience considering the amount of times played within hours but still very entertaining nonetheless. Graphics look incredible especially given what devices people may be viewig content on - top job guys! Definitely deserving 9 stars rating.
This mobile game offers much more than expected from either sides involved; developers and players alike.. With beautiful graphics, well thought background stories and plenty of engaging tasks to do in between,Flufblefox takes advantage off both smartphones’ high graphical processing abilities (iPhoness?)and Android phones.’s vast array of sensors. Not to forgetting its many touch screen gesturizations either. Overall pretty great stuff going forward specially when combined underneath the hood.
With its quirky style designs, colorful landscapes, friendly villagers and enchantingly composed score tracks,it doesn't take long after startingupforone toremember oneselfas an eager hero settoexplorea wide rangeofinterestingsubjectsbothonthehorizonandbeneathyourfeet. And despite its simple concept(whichdoesmakeitlesslikemanyotherendless running runner),the gaming experieneveninthefirst houris alsolife threateninh,and thus requiring careful considerationfromplayerswhileexplorinbg,evenminute-by– minute. Thats whywhenplayedagain and againthere is always something newtofind,either intheformoffreshnewcharacterstoaccompanyyou

setzer22 (approved):
Looks good! Thanks a lot :)

PR #66, opened 2023-03-23
fix 65B model 

jempabroni:
Import fixes for magic (https://github.com/setzer22/llama-rs/issues/59#issuecomment-1479910799). 

Fix 65B model error (https://github.com/setzer22/llama-rs/issues/65).

A better fix for the memory allocation issue is probably possible, but this works without any impact on memory for any model (tested 7B: 4G, 13B: 5G, 30B: 20G, 65B: 39G).

### DIFF
--- a/llama-rs/src/lib.rs
+++ b/llama-rs/src/lib.rs
@@ -147,7 +147,6 @@ type Token = String;
 pub struct Vocabulary {
     /// Maps every integer (index) token id to its corresponding string
     mapping: Vec<Token>,
-    score: Vec<f32>,
 }
 
 #[derive(serde::Serialize)]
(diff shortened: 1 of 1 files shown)


setzer22:
I think this implements the same changes as #61? :sweat_smile:

llama-rs/src/lib.rs
```diff
@@ -911,7 +911,7 @@ impl Model {
             f16_: _,
         } = self.hparams;
 
-        let mut buf_size = 512 * 1024 * 1024;
+        let mut buf_size = 1024 * 1024 * 1024;
```
setzer22, line 1088: Let's add an explanatory comment so others find this a bit less confusing :thinking: 
```suggestion
        // For the first run, we need to guess a maximum buffer size so we can measure
        // the actual memory consumption of the temporary ggml context.
        let mut buf_size = 1024 * 1024 * 1024;
```

setzer22 (approved):
Thanks! I just added a little explanatory comment, we should be good to merge

PR #69, opened 2023-03-24
Strip off last \r\n in the prompt file.

### LINKED ISSUE #60: Strip trailing newline in prompt file
Many, probably most, editors include a trailing newline at the end of a text file, which is not what you want if the prompt isn't a complete sentence or paragraph. This issue was fixed in llama.cpp.

KerfuffleV2:
Naturally also works if it's just `\r` or just `\n`. (However `\n\r` would result in `\n`. If you put weird stuff like that in your prompt file, you deserve what you get.)

Closes #60

### DIFF
--- a/llama-cli/src/main.rs
+++ b/llama-cli/src/main.rs
@@ -128,7 +128,19 @@ fn main() {
 
     let prompt = if let Some(path) = &args.prompt_file {
         match std::fs::read_to_string(path) {
-            Ok(prompt) => prompt,
+            Ok(mut prompt) => {
+                // Strip off the last character if it's exactly newline. Also strip off a single
+                // carriage return if it's there. Since String must be valid UTF-8 it should be
+                // guaranteed that looking at the string as bytes here is safe: UTF-8 non-ASCII
+                // bytes will always the high bit set.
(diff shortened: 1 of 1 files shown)


philpax:
Seems reasonable, but is there a reason not to use [str::trim_end](https://doc.rust-lang.org/std/primitive.str.html#method.trim_end) or [str::trim_end_matches](https://doc.rust-lang.org/std/primitive.str.html#method.trim_end_matches)?

KerfuffleV2:
@philpax Because those would strip off _all_ the trailing newline/carriage return characters, making it 1) impossible to actually use a prompt from a file that ends with a newline and 2) different behavior from llama.cpp which is probably what people will expect it to function like.

I care more about the first one there personally since I think that may be something people will actually want/need to do.

philpax:
Fair enough!

KerfuffleV2:
By the way, I hope that comment wasn't too blunt. With the numbered items it does kind of read like I was lecturing you or something. So it may have come off stronger than I intended (I'm just generally a pretty direct person though).

philpax:
Oh no, don't worry - you explained it just fine. 'sall good :)

Issue #56, opened 2023-03-22
Hi, is there any plans for word embeddings?


hlhr202:
Noob here, excuse me for my stupid feature request.
I noticed that someone in llama.cpp is working on word embedding from hidden layers. I m just asking is there any possibility to implement an embedding mode for llama-rs? thx

What i found is [this](https://github.com/StrikingLoo/llama.cpp/commit/e4524da1a173d139c8702a8db5589926d797ebd9) commit

hlhr202:
Hi, I would like to add llama.cpp PR here for reference. just noticed they merged the embedding function
https://github.com/ggerganov/llama.cpp/pull/282/files

setzer22:
Hi @hlhr202! :wave: 

Thanks for bringing this to our attention :smile:  The code here doesn't look hard at all to port! We will add it to the repo since it makes sense to have a way for people to extract embeddings.

But I'd like to understand (just to satisfy my curiosity). Why are the LLaMA embeddings useful? Is this the same thing as regular word embeddings from any other model? That is, capture the semantics of a word as a vector to allow computing similarity metrics? Do you have a use case for extracting the embeddings that would help us understand the possibilities better? :smile:

Not saying this is a requirement for the PR, I just want to learn if there are different use cases for this that I'm not aware of.

setzer22:
Please check out #72. I implemented some code to extract embeddings, but we still need to validate if the results are correct, and how to best expose this to our different levels of API.

hlhr202:
> Hi @hlhr202! 👋
> 
> Thanks for bringing this to our attention. The code here doesn't look hard at all to port! We will add it to the repo since it makes sense to have a way for people to extract embeddings.
> 
> But I'd like to understand (just to satisfy my curiosity). Why are the LLaMA embeddings useful? Is this the same thing as regular word embeddings from any other model? That is, capture the semantics of a word as a vector to allow computing similarity metrics? Do you have a use case for extracting the embeddings that would help us understand the possibilities better? 😄
> 
> Not saying this is a requirement for the PR, I just want to learn if there are different use cases for this that I'm not aware of.

yes, computing semantic similarity is quite useful in many cases. it allow us to search sentences in similar semantic by using natural language query.
 btw i will help to simply verify the pr and quickly merge into my llama-node.

hlhr202:
> Please check out #72. I implemented some code to extract embeddings, but we still need to validate if the results are correct, and how to best expose this to our different levels of API.

[...]

Issue #31, opened 2023-03-17
Non-`ggml` backend


philpax:
This has been a topic of some discussion in #4 and on the Discord, so I figured I'd document our initial findings so far.

We would like to switch away from `ggml` at some point so that we can remove the C compiler dependency, and enable running on other types of devices (namely the GPU).

Our primary candidate for a Rust-native ML/tensor backend is [burn](https://github.com/burn-rs/burn), which is a flexible deep learning framework that supports multiple backends (including ndarray and torch).

Unfortunately, it doesn't support the two formats we need: `f16` (original weights) and `q4_0`/`q4_1` (quantized weights). Adding these to the `ndarray` backend should be viable, but getting it right and working optimally (i.e. similar to `ggml`'s optimisations for those datatypes) will take some time.

Torch does support `f16` on the GPU only, but `burn`'s `torch` backend doesn't yet. That shouldn't be too hard to rectify, but the primary problem there is actually just testing: the 7B weights are 14GB, which is difficult to make work with most consumer GPUs.

So we're in a bit of a pickle - there are three options available, all of which will require some work, and all of which have individual drawbacks:

1. Quantize the model to standard `uint8` and use `ndarray`/`torch` backends. This is the least work (at least in theory), but uint8 quantization performs worse than either `f16` or `q4`, from what I've heard.
2. Add support for `f16` to `burn`'s `torch` and/or `ndarray` backends. The `torch` backend should be relatively straightforward as Torch already supports `f16` for GPUs, but it will be very hard to test with most of our machines. Adding support to `ndarray` for CPU inference shouldn't be impossible either (especially if we just convert to `f32` for every operation), but it will be difficult to make it performance-optimal.
3. Add support to `q4_0/1` to `burn`'s `ndarray` backend. This is the option that will give us the most parity with the current implementation (assuming the majority of our users are using `q4` weights), but it has the same performance-optimality issue as `f16` on the CPU (every cumulative operation, like matrix multiplication and such, will need to be specialised). Additionally, there is no way to natively store a 4-bit element, so there's no guarantee that this will be space-optimal (e.g. we can't assume that `ndarray` and `rustc` will remap `[[bool; 4]; N]` to `[u8; N/2]`).

This is summarised in the following table:

|  | `uint8` | `f16` | `q4` |
| - | - | - | - |
| `ndarray` | Yes, but at noticeable quality loss | Requires semi-significant implementation work | Requires significant implementation work |
[...]

PR #72, opened 2023-03-24
Embedding extraction

setzer22:
Implements #56.

I ported the llama.cpp code to allow extracting word embeddings and logits from a call to `evaluate`. I validated this using an `ad_hoc_test` (currently hard-coded in `main`) and results seem to make sense: The dot product for different embeddings is higher, the more similar the two words are, which is exactly how embeddings should work.

This serves as a proof of concept, but we need to discuss the API before we can merge. Currently, I added an `EvaluateOutputRequest` struct, so we can expand this in the future, allowing retrieval for other interesting bits of the inference process, but these values are not easily obtainable using the regular APIs (i.e. `feed_prompt`, `infer_next_token`). I'm not sure if that's a problem: Are we ok with users having to drop down to the lower level `evaluate` function when they need to retrieve this kind of information?

On a different note, I would really like for someone with a bit of understanding to validate that the results here are correct. Perhaps @hlhr202 can shed some light there?

Finally, should we consider exposing this to `llama-cli` at all?


llama-cli/src/main.rs
```diff
@@ -314,3 +317,81 @@ fn main() {
         }
     }
 }
+
+fn ad_hoc_test() {
```
setzer22, line 321: This function will be removed before we merge. It's just here so others can validate my results and test things.

llama-cli/src/main.rs
```diff
@@ -314,3 +317,81 @@ fn main() {
...
+        "I have just adopted a cute ",
+        |_| Ok(()),
+    );
+    let mut output_request2 = EvaluateOutputRequest {
+        all_logits: None,
+        embeddings: Some(Vec::new()),
+    };
+
+    // Try other words: 'dog', 'cat', 'potato', '$' -> To see decreasingly lower dot product values.
+    let dog2 = model.tokenize(&vocab, "dog", false).unwrap();
```
setzer22, line 379: What I'm doing here is feeding the following two sentences through the transformer:

- "My favourite animal is the dog"
- "I just adopted a cute dog"

Afterwards, I retrieve the embeddings for the last token (dog), and compute their similarity with a simple dot product.

Then, I tried changing the second sentence from 'dog' to 'cat', 'potato', '$' respectively, and the semantic similarity dropped accordingly, with $ ranking the lowest.

llama-rs/src/lib.rs
```diff
@@ -1065,11 +1077,15 @@ impl Model {
     }
 
     /// Evaluates the transformer.
+    ///
+    /// The provided `output_request` struct lets you specify which additional
+    /// data you are interested in fetching from the transformer. Setting a field to a `Some` value will fill the
```
KerfuffleV2, line 1082: I think you might have run out of context tokens here. :(

llama-cli/src/main.rs
hlhr202: @setzer22 will feed prmpt before eval has different embeddings compared to eval all tokens together?

llama-rs/src/lib.rs
```diff
@@ -1296,6 +1316,30 @@ impl Model {
...
 
+        // Extract logits
+        if let Some(all_logits) = &mut output_request.all_logits {
+            all_logits.resize(n_vocab as usize * n, 0.0);
+            // SAFETY: Data can be read (properly aligned, initialized, data
+            // will not be mutated or otherwise aliased while the slice lives),
+            // and we're not reading past the end of the slice.
+            assert_eq!(input_layer.nelements(), n_vocab * n as i32);
+            let logits_slice = unsafe {
+                slice::from_raw_parts(input_layer.data() as *const f32, n_vocab as usize * n)
```
philpax, line 1327: Could you add a function to `Tensor` for getting the data out as a slice? This is what I did:
https://github.com/philpax/ggllama/blob/main/src/ggml.rs#L105-L114

philpax:
LGTM once the other review feedback's sorted out. 

For exposing it from the CLI, I'm not sure... people *might* use it as a process in a CLI pipeline (get the embeddings of two texts and then comparing them), but I'm not sure what that would look like or how people would do that. (What output format would we use?)

Unless someone can suggest a "standard" output format for this, I'd suggest leaving it out for now and figuring it out later.

hlhr202:
> LGTM once the other review feedback's sorted out.
> 
> For exposing it from the CLI, I'm not sure... people _might_ use it as a process in a CLI pipeline (get the embeddings of two texts and then comparing them), but I'm not sure what that would look like or how people would do that. (What output format would we use?)
> 
> Unless someone can suggest a "standard" output format for this, I'd suggest leaving it out for now and figuring it out later.

It would be just nice for me to expose such get embedding function as in crate library. actually i do not care much about cli exposing. what I v seen llama.cpp they provide a parameter --embedding for output purpose. but they still did not find out a way to expose it though. thats why i still cannot get the embedding from their cli currently.
I only test a few cases with comparing to openai's embedding. should be some difference, but i think that is caused by different model.

KerfuffleV2:
@hlhr202 The CLI is just a consumer of the library crate, so when using the library you'll be able to get the embeddings.

hlhr202:
> @hlhr202 The CLI is just a consumer of the library crate, so when using the library you'll be able to get the embeddings.

yes absolutely, cuz i m porting llama-rs to llama-node, so i just need library pub function exposing.
it doesnt make sense to expose embeddings in cli anyway.

llama-rs/src/lib.rs
setzer22: Good point!

But, there's a caveat: I don't think this function can be marked as safe. Once something is in a slice, it has to adhere to Rust's aliasing model. The way tensors are implemented, a tensor has to behave like a cheap-to-copy pointer that supports aliasing (this is used in several places of the inference code). This is why tensors have a `share()` function to make this aliasing explicit.

So even if your function tie the lifetime of the tensor to the output slice, there could be many other aliased references to that same tensor through which you could obtain another mutable slice to the same memory. And that's unsound.

What we can do, is have a similar API but mark it as `unsafe` and document the safety requirements for it. But I honestly think it todesn't make sense to try to build an idiomatic safe rust wrapper to ggml. It is a C library built around the idea of aliasing pointers and "being careful" not to drop a context before some other context uses it :sweat_smile:. Even if Rust lifetimes could somehow capture that, the loss in ergonomics would be a high price to pay. One of the main design goals in the original design was to make the code as easy to port as possible. This includes any changes to the tensor code that may happen in the future, like #68.

Any efforts towards making ggml more Rusty would be better directed towards #31.

llama-cli/src/main.rs
setzer22: @hlhr202 The embeddings wouldn't be affected, but you shouldn't call `evaluate` with the whole prompt like that for a couple reasons:

- A call to evaluate runs all the tokens you give it as a batch, meaning it requires increased memory usage. For very long prompts, this could become very expensive.
- The output value will return the output embeddings for every token that you fed through `eval`. This means you would be retrieving a lot more embedding data than for just the word "dog".

llama-rs/src/lib.rs
setzer22: I just realized we already had a similar function for this (`read_data`), so I used that one instead :+1:

setzer22:
I already addressed the review feedback and removed the ad-hoc test code. So I take it a good plan now would be to merge this as-is and have embedding extraction as a low-level feature of llama-rs, but simply not expose it to the CLI?

llama-rs/src/lib.rs
philpax: @setzer22 fyi

philpax (approved):
LGTM, ready to merge after the comment's fixed

KerfuffleV2:
Since I added the `--dump-prompt-tokens` option, you can probably guess I like exposing information. :) I know people asked about being able to show the embeddings with llama.cli, so it does seem like there's some kind of demand for it in a CLI.

philpax:
If there's demand, I'm happy to do so - just not sure what the output format should be. JSON array or newline-delimited floats?

KerfuffleV2:
Is it a lot of data? You could probably just print in the normal Rust debug format which should look like a comma separated list if it's in a `Vec` or something. That should be pretty easy to transform to other formats without need to write extra code or pull in dependencies.

llama-rs/src/lib.rs
setzer22: Oops, missed that one, should be fixed now :+1:

setzer22:
> Is it a lot of data?

It is quite a lot of data for comfortably printing to stdout honestly. It's 4096 floats per token.

KerfuffleV2:
Ahh, then seems like it probably isn't worth even bothering to add to the CLI right now unless someone comes here and requests it. Or they could probably just write their own little utility to load a model, feed a prompt and print out embeddings however they wanted.

llama-cli/src/main.rs
hlhr202: @setzer22 I just understand your comments here. this means we can only extract embeddings for a single part of words (which may also have hidden information mixed with context of the whole sentence). that should a little bit different with OpenAI's embedding function. what i understand for openai's embedding, is for the whole sentence but at the same time returned in a fixed size of tensor... that is quite beyond my knowledge though

hlhr202: well I guess I might find possible ways to implement such 'sentence embedding', I will try add some special end token and extract the hidden layer once the end token evaluated. not sure if it works, but it must worth a try.

rpbrokaw:
> Ahh, then seems like it probably isn't worth even bothering to add to the CLI right now unless someone comes here and requests it. Or they could probably just write their own little utility to load a model, feed a prompt and print out embeddings however they wanted.

I would love that in the CLI! Perhaps with a parameter that specifies an output file. I need the embeddings to build a vector database based on some local files. Any chance you could take a look? It has been many years since I programmed C/C++.

hlhr202:
> > Ahh, then seems like it probably isn't worth even bothering to add to the CLI right now unless someone comes here and requests it. Or they could probably just write their own little utility to load a model, feed a prompt and print out embeddings however they wanted.
> 
> I would love that in the CLI! Perhaps with a parameter that specifies an output file. I need the embeddings to build a vector database based on some local files. Any chance you could take a look? It has been many years since I programmed C/C++.

vector is around 4096 length for one token, not very well printed in CLI. I guess you need to call in through rust api.

setzer22:
I'm open to adding a way for the CLI to output embeddings if people find this is an interesting use case. The main blocker here is that the use case is not clear to me and thus I can't figure out the right API and output format.

What we need here, is someone who understands how embeddings in an LLM like LLaMA work, and has a clear use case for extracting them and can tell us how would they expect an API like this to work. If anyone wants to open an issue with a clear description of what we need to provide, I'd be happy to add an implementation :slightly_smiling_face:

hlhr202:
@setzer22 I have made a new embedding extraction example. can check it here https://github.com/hlhr202/llama-node/blob/develop/packages/core/example/semantic-compare/compare.py
I noticed that llama.cpp they use "\n" as end token, so I also do the same. it is quite close to openai's text-embedding-ada-002 result

turbo:
I'm working on a large dense-vector embedding database (about 2 million data points from books), which is currently using OpenAI's Ada embeddings (~1600 dimensions). I can do a comparison of performance between those and the 4k LLaMa embeds if needed.

> as a clear use case for extracting them and can tell us how would they expect an API like this to work

From an ops perspective, ideally one could provide a batch input and get a batch output (just like OpenAI's API) via CLI. The format doesn't matter much - it can be JSONL or a binary format. I'd personally recommend sticking to those two, since that is supported by most VSS databases (e.g. Redis RediSearch).

merlinvn:
> o adding a way for the CLI to output embeddings if pe
My use case here is if you have a sets of documents, and if you can get the embeddings of those documents, whenever a new question comming in, you can embed the new questions and find the most relevant documents to send a long with you prompt. So basically you can have a natural Q&A chat bot based on your own data.

Issue #63, opened 2023-03-23
Explain differences from llama.cpp in README


philpax:
`llama.cpp` now has a C interface, so we could theoretically switch to using it directly.

However, we don't want to do this for a few reasons:

- You still need a C++ compiler, which complicates deployment to other platforms
- Rust makes the code easier to work with
- We want to make ggml an optional backend in future (#31)

CSRONOML:
I don't understand why `llama.cpp` wasn't created using Rust in the first place (like a `llama.rs`), what advantage did the author see in using C++ instead of Rust?

KerfuffleV2:
He hasn't seen the light yet, that poor lost soul.

The real answer is probably because C++ is what he knows/is experienced with and also had already written stuff like GGML and Whisper in C++ already so he could just reuse a lot of it.

PR #23, opened 2023-03-16
:memo: add build instructions for fedora

sylvain-reynaud:
Closes #22

## How to test it ?

`docker run -it fedora bash`

Then

```bash
ls /usr/lib/        #no gcc
dnf groupinstall "Development Tools" "Development Libraries"
ls /usr/lib/gcc/x86_64-redhat-linux/12/include/        #installed
git clone https://github.com/sylvain-reynaud/llama-rs
cd llama-rs/
curl https://sh.rustup.rs -sSf | sh
source "$HOME/.cargo/env"
cargo build --release        #still fails
CPATH="/usr/lib/gcc/x86_64-redhat-linux/12/include/" cargo build --release        #success
```

PR #73, opened 2023-03-25
docs: update README with info + instructions

philpax:
This PR reformats the README (making it markdown-lint compliant), adds more instructions, and clarifies the differences between this and `llama.cpp`.

It should address #23, #50 and #63.


setzer22 (approved):
Very nice! Thanks for taking care of this :smile: 

Hahaha, I see you kept the original Q&A around. I believe there has to be a place in OSS for humor, and I like keeping those for the time being. But as the project shifts into more serious territory, I guess we will end up removing the original joke questions. Perhaps at a later PR, when we have more things to add to the Q&A :+1:

Issue #57, opened 2023-03-22
Publish to crates.io


philpax:
We should publish the crate and its associated applications to `crates.io` (potentially bringing llamacord etc into a GitHub organization, too). Here's what I think's blocking:

- [x] Move `ggml-rs` to its own publishable crate, add safety warnings
- [x] Fix warnings
- [x] Add deny public docs, document everything
- [x] Get all pending PRs in

Anything else I'm missing, @setzer22 ?

philpax:
Once #81 is done, setzer can publish

vv9k:
This issue has been closed but I don't see a working version on crates.io . Currently there is a version 0.0.0 that contains an empty `main.rs` file printing "hello world". Is the release still to be done or am I looking in the wrong place?

https://docs.rs/crate/llama-rs/latest/source/src/main.rs

end-me-please:
love to see how fast development of this stuff is going
how seamlessly this thing works together with llama-node (compared to the less "clean feeling" llama.cpp) makes me want to start rewriting random stuff in rust (🚀)

especially loving the low number of dependencies compared to the p*thon AI stuff that is used everywhere else

philpax:
We have a RC1 published for all crates 🚀 

My plan is to publish 0.1.1 (I made the mistake of publishing `llm 0.1.0` for reservation) tomorrow.

philpax:
Took a bit longer than expected (had to sneak in an API change), but 0.1.1 is live on crates.io 🎉

Issue #40, opened 2023-03-18
Convert `quantize.cpp` to Rust


philpax:
Split this off from #21 as it's a separate issue.

This should be relatively straightforward - it reads in the original `ggml` model, runs the quantization functions over the data, and writes it out to disk.

The exciting possibility is for parallelisation 👀 - all you should have to do is scan through the file to determine the tensor boundaries, then build an iterator from it and feed it to `rayon`. It would be a *huge* improvement over the C++ version, and it would be practically free!

FloppyDisck:
If theres nobody working on this I could tackle it in the week

FloppyDisck:
Is there currently a way to convert models to ggml format? Im close to getting quantize into a working demo and was wondering if this should also be ported for the PR

philpax:
Nope, that still requires the original Python code. If you want to tackle #21 as well, that would be awesome!

FloppyDisck:
I have no problem working on it after I finish this issue. 

Another question, how should the feature be used? Should it be another argument in the cli app?

philpax:
Hm, just do the simplest possible thing for now and we'll figure out a new CLI. There's several changes landing to the CLI soon, so we should avoid doing anything complicated until it's entirely resolved.

PR #74, opened 2023-03-25
WIP: Refactor Cli

hhamud:
Refactor and clean up of the Command-line interface.

This PR aims to decouple prompt generation from model generation and make it easier to expand upon in the future.

### TODO:
*  [ ] Decouple prompt generation from model loading and 
* [ ] Separate CLI UI from business logic 
* [ ] Create separate modes for the CLI


philpax:
You'll probably want to wait until the other PRs land

hhamud:
> You'll probably want to wait until the other PRs land

Which ones in particular are you thinking of?

philpax:
#41, #68, #72, and once those are in I'm going to do some cleanup work in preparation for #57

philpax:
Getting closer to being able to unblock you on this. In the meantime, you may be interested in looking into revising the CLI parameters to accommodate the different operational modes we have here; I suggest using Clap subcommands.

As an example of how you can share those parameters, here's how I do it in one of my other projects: https://github.com/AmbientRun/Ambient/blob/main/app/src/cli/mod.rs#L7-L100

hhamud:
> Getting closer to being able to unblock you on this. In the meantime, you may be interested in looking into revising the CLI parameters to accommodate the different operational modes we have here; I suggest using Clap subcommands.
> 
> As an example of how you can share those parameters, here's how I do it in one of my other projects: https://github.com/AmbientRun/Ambient/blob/main/app/src/cli/mod.rs#L7-L100
> 
> (This should also help with offering a mode for #40)

When I started repl-mode was about merged in and I thought about creating an interactive-mode for the "chat-like" app but I can see there are more on the way.

philpax:
#83's refactored the CLI a fair bit - do you think there should be more work in that area?

hhamud:
> #83's refactored the CLI a fair bit - do you think there should be more work in that area?
There were a few additional things that I wanted to do after #85 get's merged that would further change the way models are loaded into the CLI.

philpax:
Hm, should I close this PR and we'll revisit once #85's in?

PR #76, opened 2023-03-26
Keep context in repl

LukeAngove:
The current repl implementation drops the context with every instance of the prompt. This means the user can't refer to past information, and the model can't use past information to figure out the current context.

This PR also splits out the initial prompt and the per comment prompt, as there's no need to repeat the initial prompt multiple times, and it fills up the context faster. It also seems to confuse the model a lot.

Note the responses when doing this can become a bit odd pretty quickly (lots of #LotsOfWordsInARow and emoji), but I think the code change is more correct for the intent of the repl use case.

The initial prompt given needs a bit of work, but it seems to function at a basic level.

Tested with:
`cargo run --release -- -F examples/initial/custom_alpaca.txt -f examples/custom_alpaca_prompt.txt --model-path ggml-alpaca-native-7b-q4.bin --repl --num-ctx-tokens 4096`


setzer22:
Hi! Thanks for sending the PR :)

One thing needs to be made clear though: The repl mode doesn't intend to be a chat application. Unfortunately we don't seem to have a good way to implement a chat-like application on top of LLaMA just yet. The instruction-tuning done for alpaca is good for replying to individual requests, not for keeping a long-term conversation with lots of back-and-forth, so it will misbehave when you try to do that. The limited context window size (2048 tokens) also makes a chat-like application unfeasible overall.

Some degree of success may be possible by doing some prompt engineering on top of the raw LLaMA models. But with those being raw models with no fine-tuning, we can expect performance to be quite bad.

So, overall, I don't think we should change the behavior of `--repl` mode. The main use case behind that is to allow running multiple queries against alpaca without having to unload and load the model all the time.

A PR like this is still welcome, but I would prefer if we set the right expectations for users and create a separate `--chat-experimental` flag for this. The naming should make it clear that this is not going to work well with the widely available models and that requires some further tweaking.

LukeAngove:
I've made some updates that retain the old behaviour by default, but if 'chat_rules' are set, then state is maintained until the buffer overflows. I've also added a print statement to the start of the 'repl' environment to clarify which mode is in use. 

Let me know what you think. Happy to make any further changes if desired.

philpax (approved):

LukeAngove:
In resolving the conflicts with main, I think the merged PR https://github.com/rustformers/llama-rs/pull/91 actually adds this feature (possibly inadvertently), and without an initial prompt. It moves the creation of the session outside of the loop, meaning that the same session is kept for each iteration in the REPL. Let me know if I'm mistaken.

philpax:
#83's implemented this - thank you for pointing this out as an issue 🙂

Issue #80, opened 2023-03-26
Use `RMSNorm` for normalization


philpax:
Upstream has switched to using RMSNorm for normalization, which is more accurate to the original implementation: https://github.com/ggerganov/llama.cpp/issues/173

We didn't do this at first because there seemed to be some issues, but those seem to have been resolved.

Issue #62, opened 2023-03-23
Update to latest llama.cpp


philpax:
There have been quite a few changes since our last major sync: https://github.com/ggerganov/llama.cpp/compare/904d2a8d6acd667c9633138d45a361d40fbf76d0..HEAD

(There may be others we haven't accounted for in the inferencing code).

Need to do a more precise breakdown, but

- #59 and #61

iacore:
I met the `InvalidMagic` error again, after #61. Is there another breaking change in ggml format?

philpax:
Yep, #93.

Issue #48, opened 2023-03-21
Make `InferenceSession` `Clone`-able


philpax:
In one of my test applications, I use an `InferenceSession` to load in a prompt that I later reuse. However, I realised while doing this that you can't actually clone an `InferenceSession` in memory (and I think it *should* be possible?), so I had to serialize the session to a `Vec<u8>` and rehydrate it when I needed to infer from it.

I think this should be easy enough to fix, but we should check that there aren't any weird assumptions that we're violating if we do so. (I assume this would also allocate another `ctx`, but that should be fine)

setzer22:
Yup, I don't see any problems here (other than this just hasn't been implemented yet) :smile: 

This might require some careful handling of the underlying ggml context. Make sure a new context is allocated and any tensor data is copied over to the new context. For this, you might need some C-like pointer fiddling and maybe expose a few more GGML functions. Simply cloning the pointers would result in the wrong behavior, and most likely UB.

Issue #44, opened 2023-03-19
API footgun: `infer_next_token` still works after end of text


philpax:
In llamacord, I have some logic that calls `infer_next_token` in a loop. Unfortunately, I didn't check for EOT - so the code would keep generating tokens and producing (fascinatingly well-structured) garbage. I *think* we should probably check if the last token is EOT and return an error? If you feed it a prompt, the EOT would no longer be the last token, and you should be able to infer without issues. (I wonder if A[EOT]B infers differently to AB...)

setzer22:
I'd say being able to infer beyond eot is a feature some might want, even if it's just to run some experiment to see what would happen. But I'm OK with making the API harder to misuse, as long as it's still possible to request inference for a new token after EOT.

Returning some sort of "EOT" error in `infer_next_token` sounds like it might be what we want?

> I wonder if A[EOT]B infers differently to AB...

Yes, for sure. EOT is an important token and the transformer will interpret both strings completely differently :+1:

philpax:
Yeah, that's what I was thinking. It wouldn't allow your use-case, though: if you returned an EOT error, you wouldn't be able to get the inferred token back.

philpax:
Discussed this on the Discord - it makes much more sense to return the EOT token itself as an error, and then let users continue inferring if they want.

PR #81, opened 2023-03-26
Prepare for release

### LINKED ISSUE #57: Publish to crates.io

philpax:
Mostly housekeeping.

Closes #57.

Still to do:

- [x] Document `llama-rs`
- [x] #62 
- [x] #18 
- [x] #80
- [x] #48
- [x] #44
- [x] Fix performance on Apple Silicon


philpax:
This is good to go aside from the AS fix (need to test on work laptop).

Something I noticed is that `InferenceParameters` contains `play_back_previous_tokens` (only relevant to `inference_with_prompt`) and `inference_with_prompt` takes `maximum_token_count` as a parameter - we should consider moving these into a specific struct for parameters for `inference_with_prompt` only, like
```rust
    pub fn inference_with_prompt<E: std::error::Error + 'static>(
        &mut self,
        model: &Model,
        vocab: &Vocabulary,
        inference_params: &InferenceParameters,
        inference_with_prompt_params: &InferenceWithPromptParameters,
        prompt: &str,
        rng: &mut impl rand::Rng,
        callback: impl Fn(&str) -> Result<(), E>,
    ) -> Result<InferenceStats, InferenceError> {
}
```

We could also potentially move the `&mut impl rand::Rng` into `InferenceParameters`. I'm not sure if this is a good idea yet, but it should let us simplify the number of arguments some more.

Is there a context in which you would use the `Vocabulary` of a model without the model itself? I'm leaning towards moving that into `Model`, too.

KerfuffleV2:
> We could also potentially move the `&mut impl rand::Rng` into `InferenceParameters`. 

You'd have to add a lifetime to the struct and maybe it would need to be mutable also, so probably not worth it.

hhamud:
I'm on apple silicon and I haven't felt any performance issues versus the original implementation of llama.cpp in C. What are the performance issues?

llama-rs/src/lib.rs
```diff
@@ -98,14 +105,39 @@ impl InferenceSession {
             .saturating_sub(self.params.repetition_penalty_last_n)..]
     }
 }
+impl Clone for InferenceSession {
+    fn clone(&self) -> Self {
+        let context = ggml::Context::init(self.memory_size);
+        let memory_k = context.new_tensor_1d(
+            self.memory_k.get_type(),
+            self.memory_k.get_ne()[0].try_into().unwrap(),
```
setzer22, line 113: there's `nelements()` too which might be a bit more succint

setzer22 (approved):
Thanks a lot for the changes! :)  Everything looks good, there's just a very minor nitpick comment from my end.

Also, why does `ggml.rs` appear as a removed file? The code seems to still be using it, and I don't see where that code has moved to :thinking:

philpax:
> Thanks a lot for the changes! :) Everything looks good, there's just a very minor nitpick comment from my end.
> 
> Also, why does `ggml.rs` appear as a removed file? The code seems to still be using it, and I don't see where that code has moved to 🤔

It's moved to `ggml/src/lib.rs`. Not sure why Git can't detect the move.

setzer22 (approved):
LGTM! :smile:

Issue #21, opened 2023-03-16
Directly load `pth`/PyTorch tensor model files


philpax:
At present, `llama.cpp` contains a Python script that converts `pth` to `ggml` format.

It would be nice to build it into the CLI directly, so that you can load the original model files. The original Python script could also be converted to Rust, so that we have a fully-Rust method of converting `pth` to ggml models.

philpax:
re loading `pth`: [serde-pickle](https://crates.io/crates/serde-pickle) looks quite promising, but we would need to figure out if it can load PyTorch tensors.

philpax:
This is not complete yet. We've merged in the start of a converter, but more work is required to convert the weight.

Luckily, @KerfuffleV2's developed a Pickle parser that can handle PyTorch tensors: https://github.com/KerfuffleV2/repugnant-pickle

We should be able to use this to convert tensors to GGML format. In future, we can directly load tensors (I may separate that out into a new issue), but our focus is on loading tensors so that they can be quantised by #84 and used by `llama-cli`.

PR #83, opened 2023-03-27
WIP: Convert `pth` to `ggml`

karelnagel:
I'm still working on adding the weights to the file, rn it only adds params and tokens to the file (the md5 hash matches the llama.cpp generated file without the weights). And also no quantizing.
Also added `generate` and `convert` subcommands to the cli as well.
Let me know if anything needs changing 🙂

Resolves https://github.com/rustformers/llama-rs/issues/21

### DIFF
--- a/llama-cli/src/cli_args.rs
+++ b/llama-cli/src/cli_args.rs
@@ -6,7 +6,27 @@ use once_cell::sync::Lazy;
 
 #[derive(Parser, Debug)]
 #[command(author, version, about, long_about = None)]
-pub struct Args {
+pub enum Args {
+    #[command(name = "generate")]
+    Generate(Generate),
+
+    #[command(name = "convert")]
+    Convert(Convert),
+}
+
+#[derive(Parser, Debug)]
+pub struct Convert {
+    /// Path to model directory
+    #[arg(long, short = 'd')]
+    pub dir: String,
+
+    /// File type to convert to
+    #[arg(long, default_value_t = false)]
+    pub f32: bool,
+}
+
+#[derive(Parser, Debug)]
+pub struct Generate {
     /// Where to load the model path from
     #[arg(long, short = 'm')]
     pub model_path: String,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "ahash"
+version = "0.3.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e8fd72866655d1904d6b0997d0b07ba561047d070fbe29de039031c641b61217"
+
+[[package]]
 name = "aho-corasick"
 version = "0.7.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -10,6 +16,12 @@ checksum = "cc936419f96fa211c1b9166887b38e5e40b19958e5b895be7c1f93adec7071ac"
 dependencies = [
  "memchr",
 ]
+
+[[package]]
+name = "autocfg"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "bincode"
@@ -31,6 +43,12 @@ name = "bytemuck"
 version = "1.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "17febce684fd15d89027105661fec94afb475cb995fbc59d2865198446ba2eea"
+
+[[package]]
+name = "byteorder"
+version = "1.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "14c189c53d098945499cdfa7ecc63567cf3886b3332b312a5b4585d8d3a6a610"
 
 [[package]]
 name = "cc"
@@ -72,7 +90,7 @@ dependencies = [
  "proc-macro-error",
  "proc-macro2",
  "quote",
- "syn",
+ "syn 1.0.109",
 ]
 
 [[package]]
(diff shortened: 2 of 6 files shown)


karelnagel:
I tried using `serde_pickle`, but it panics every time I try to load the file 
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Eval(Unsupported('P'), 1)', llama-rs/src/convert.rs:152:79
```
`serde-pickle` might not work for torch files according to these: https://github.com/guillaume-be/rust-bert/issues/12 , https://github.com/LaurentMazare/tch-rs/issues/171. 
Other option would be to use https://github.com/LaurentMazare/tch-rs but this would require installing c++ libtorch on the system 🥲. I will try to research more tomorrow.

philpax:
Nice one! Much appreciated - hope you can figure out what's going on with `serde_pickle`. Would really like to avoid having a dependency on Torch if possible.

karelnagel:
I had no luck with serde_pickle, running it like this: 
Screenshot 2023-03-30 at 10 54 09

And also tried to use safetensors deserialize but panics with `HeaderTooLarge`, but idk if I used it correctly. 
I can't work on this for some time, so if anyone wants to continue from here on, then go ahead 🙂

setzer22:
> I'm tempted to clean this up and merge it in but with the functionality disabled, so that we can have the base functionality in and we can develop it as we go.

I'd say this is reasonable :+1: There's some work in this PR and leaving it here for too long means it will end up diverging too much.

philpax:
I'm going to update this to the latest version, hide the CLI version for now, and merge it in - we can then work on our own parser for the tensors when we have some time 🚀

philpax:
Things that I'll fix:

- [ ] Remove `get_n_parts` (look at the directory like the load code does)
- [ ] Split out the REPL as a separate CLI mode
- [ ] Update the README
- [ ] Make it clear what the `f32` parameter does (enum?)
- [ ] Use a `Result` type for the conversion process
- [ ] Make `convert` an optional feature, including its dependencies
- [ ] Clean up `main.rs` in general
- [ ] Change `Vocabulary::from` to a normal function (obscures the load operation)

karelnagel:
Nice! All good for me

KerfuffleV2:
Actually, I just remembered something that might help you with your pickle problem: https://github.com/BlinkDL/ChatRWKV/issues/40#issue-1627748575

That's some example code for manually loading a `.pth` file (they're actually ZIP files with the data uncompressed, so it's contiguous and can be read directly). From that example, you can basically just figure out what files and ranges of data in the `.pth` are associated with what without needing Torch or even having to load the entire thing.

Using this approach would still need a little helper Python script but it could do something like just scan through the tensors in the `.pth` file and collected the metadata. Only `data.pkl` is actually going to be pickled, the actual tensors are just raw data.

KerfuffleV2:
I've been experimenting with this and it's definitely not something as simple as accidentally loading the ZIP file. `serde_pickle` doesn't support the `BINPERSID` opcode, and it also can't handle stuff using `OrderedDict`. I hacked support for both of those into it and it can load the pickle, but the results aren't correct. I can't tell if it's a problem specifically to do with these changes or just because `serde_pickle` is buggy.

Also, I'd say whoever invented the pickle format should be taken out back and shot, but that's far to clean an end for someone responsible for such heinous crimes.

llama-cli/src/cli_args.rs
```diff
@@ -1,34 +1,94 @@
...
-    #[arg(long, short = 'm')]
-    pub model_path: String,
+pub enum Args {
+    #[command()]
+    /// Use a model to infer the next tokens in a sequence, and exit
+    Infer(Box<Infer>),
+
+    #[command()]
+    /// Use a model to dump the tokens for a given prompt
+    DumpTokens(Box<DumpTokens>),
```
setzer22, line 18: The description for "dump tokens is very ambiguous. I'd bring back the original:

> /// Dumps the prompt to console and exits, first as a comma-separated list of token IDs
    /// and then as a list of comma-separated string keys and token ID values.

llama-cli/src/cli_args.rs
```diff
@@ -1,34 +1,94 @@
...
+
+    #[command()]
+    /// Use a model to interactively generate tokens
```
setzer22, line 21:
```suggestion
    /// Use a model to interactively prompt it multiple times, while
    /// resetting the context between invocations
```

llama-cli/src/cli_args.rs
```diff
@@ -1,34 +1,94 @@
...
+    /// Use a model to interactively generate tokens, and chat with it
+    ///
+    /// Note that most, if not all, existing models are not trained for this
+    /// and do not support a long enough context window to be able to
+    /// have an extended conversation.
+    ChatExperimental(Box<Repl>),
 
-    /// The prompt to feed the generator
+    #[command(hide = true)]
+    /// Convert a PyTorch model to a GGML model
```
setzer22, line 33: I would document that this does not currently work. I know it's hidden, but I mean that adding a TODO telling developers that this feature is disabled, and maybe even link at this PR sounds worthwile.

llama-cli/src/cli_args.rs
```diff
@@ -137,20 +170,192 @@ impl Args {
...
-pub static CLI_ARGS: Lazy<Args> = Lazy::new(Args::parse);
+#[derive(Parser, Debug)]
+pub struct ModelLoad {
+    /// Where to load the model path from
+    #[arg(long, short = 'm')]
+    pub model_path: String,
+
+    /// A file to read the prompt from.
+    #[arg(long, short = 'f', default_value = None)]
+    pub prompt_file: Option<String>,
```
setzer22, line 235: It feels a bit weird that the prompt file is inside the `ModelLoad` subcommand.

llama-cli/src/cli_args.rs
```diff
@@ -137,20 +170,192 @@ impl Args {
...
+
+    /// Sets the size of the context (in tokens). Allows feeding longer prompts.
+    /// Note that this affects memory. TODO: Unsure how large the limit is.
+    #[arg(long, default_value_t = 512)]
+    pub num_ctx_tokens: usize,
```
setzer22, line 240: By now I'm quite convinced the limit for this is 2048. Trying to generate beyond that value starts producing garbage. Setting a lower value can be used to slightly reduce memory usage, but has no effect on inference quality.

I don't want to make 2048 a hard error because we want to be able to run larger models if someone figures out a way to fine-tune LLaMA (or another model) to extend the context window. But we should document this is the maximum for LLaMA for the vast majority of users that don't know.

I think what we should do is set 2048 as the default, so that users don't need to worry about raising this. I think the increase in memory usage is not that much. Based on some napkin math, the full context window takes ~1.6GB of Ram (4096 embd * 2048 ctx * 24 layers * 4bytes), or half of it when using F16. The the default of 500 would be around ~400MB. But these are dwarfed anyway by the model's weights (especially since this value doesn't increase in larger models). So I don't think many people are benefitting from the lower default.

KerfuffleV2:
No idea why I put so much time into this, but: https://github.com/KerfuffleV2/repugnant-pickle

You can now (hopefully) parse tensor metadata from PyTorch files in Python.

See this part for an example of what you get: https://github.com/KerfuffleV2/repugnant-pickle#pytorch

setzer22:
I have no more time to review things today, but *YES PLEASE* :sob: 

Being able to load the real weights directly would be so much nicer for users.

setzer22 (approved):

KerfuffleV2:
@setzer22 Just to be clear, what I wrote just allows interfacing with the PyTorch files and discovering the tensor metadata (what tensors exist, what types, the dimensions, where they are in the file). That will help facilitate writing something like a conversion utility without needing to involve Python or Torch.

However, if you wanted to do something like load the original non-GGML PyTorch model that would be a much more difficult task since the tensors aren't in the GGML format, may not be quantized, etc.

KerfuffleV2:
Very easy: https://github.com/KerfuffleV2/smolrsrwkv/blob/182cd3205b7a7c95571a09bcfbb954b0041e4f90/smolrwkv/src/loader.rs#L87

That function expects to get a filename + the `mmap`ed entire `.pth` file (just as an example, you can do it other ways). Based on the absolute offset + tensor shape and element size, it can calculate the range of data in the file that is associated with that tensor.

PR #84, opened 2023-03-27
Implemented `quantize.cpp`

### LINKED ISSUE #40: Convert `quantize.cpp` to Rust

FloppyDisck:
I went ahead and ported the main `quantize.cpp` file. My changes involve porting the file while keeping the internal C++ function calls intact. I have plans to port those function calls to remove ggml dependencies in a future PR though.

During the porting process, I faced some challenges as I was not familiar with how to use `Context`. As a result, I added the `half` library to handle the f16->f32 conversion. I could remove the dependency if needed but ill need some help with working with `Context`. Something to note on this is that if there are plant to move away from ggml then the `half` library will be necessary. 

Additionally, I included some print statements inside the function to mimic the original behavior of quantize.cpp. I can remove those if needed.

Currently, there is no way to access the function since I did not implement a CLI function for it. 

I am open to feedback and suggestions on how to improve this pull request.

Resolves #40

### DIFF
--- a/ggml-raw/ggml/ggml.c
+++ b/ggml-raw/ggml/ggml.c
@@ -397,6 +397,53 @@ static inline __m128i packNibbles( __m256i bytes )
 }
 #endif
 
+// method 5
+// blocks of QK elements
+// represented with a single float (delta) and QK/2 8-bit ints (i.e QK 4-bit signed integer factors)
+
+// reference implementation for deterministic creation of model files
+static void quantize_row_q4_0_reference(const float * restrict x, void * restrict y, int k) {
+    assert(k % QK == 0);
+    const int nb = k / QK;
+
+    const size_t bs = sizeof(float) + QK/2;
+
+    uint8_t * restrict pd = ((uint8_t *)y + 0*bs);
+    uint8_t * restrict pb = ((uint8_t *)y + 0*bs + sizeof(float));
+
+    uint8_t pp[QK/2];
+
+    for (int i = 0; i < nb; i++) {
+        float amax = 0.0f; // absolute max
+
+        for (int l = 0; l < QK; l++) {
+            const float v = x[i*QK + l];
+            amax = MAX(amax, fabsf(v));
+        }
+
+        const float d = amax / ((1 << 3) - 1);
+        const float id = d ? 1.0f/d : 0.0f;
+
+        *(float *)pd = d;
+        pd += bs;
+
+        for (int l = 0; l < QK; l += 2) {
+            const float v0 = x[i*QK + l + 0]*id;
+            const float v1 = x[i*QK + l + 1]*id;
+
+            const uint8_t vi0 = ((int8_t) (round(v0))) + 8;
+            const uint8_t vi1 = ((int8_t) (round(v1))) + 8;
+
+            assert(vi0 >= 0 && vi0 < 16);
+            assert(vi1 >= 0 && vi1 < 16);
+
+            pp[l/2] = vi0 | (vi1 << 4);
+        }
+
+        memcpy(pb, pp, sizeof(pp));
+        pb += bs;
+    }
+}
 
 // method 5
 // blocks of QK elements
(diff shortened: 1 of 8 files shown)


philpax:
Awesome work! I've left some feedback; it's not the most Rust-y code, but that's fine as it's a port and we can fix that up later on 🙂 

Really appreciate you doing this, it's great to get one step closer to being completely standalone 🚀

FloppyDisck:
Your comments arent showing up in the PR,

And I fully agree, im going to go ahead and fix all the clippy issues plus see if I can improve some of the logic.

Do you have any recommendations on data reading and writing? I use the same buffer multiple times since its more efficient when working on rust embedded systems so I went ahead and did the same here.

philpax:
Weird, I can definitely see the comments here and in the diff. Not sure what's happening there.

For data read/write, not sure - reusing the same buffer seems reasonable if they're semantically similar, but I'd just use a new buffer if they're not. Do you have any examples of something you'd want advice on?

philpax (changes requested):

ggml-raw/ggml/ggml.c
```diff
@@ -397,6 +397,53 @@ static inline __m128i packNibbles( __m256i bytes )
 }
 #endif
 
+// method 5
+// blocks of QK elements
+// represented with a single float (delta) and QK/2 8-bit ints (i.e QK 4-bit signed integer factors)
+
+// reference implementation for deterministic creation of model files
+static void quantize_row_q4_0_reference(const float * restrict x, void * restrict y, int k) {
```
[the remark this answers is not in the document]
FloppyDisck: I copied the bits I needed to work on this PR

llama-rs/src/quantize.rs
```diff
@@ -0,0 +1,343 @@
...
+
+                    let mut buffer = vec![0u8; (nelements * 2) as usize];
+                    finp.read_exact(&mut buffer)?;
+                    // Compute buffer
+                    for (index, chunk) in buffer.chunks(2).enumerate() {
+                        let i = u16::from_le_bytes([chunk[0], chunk[1]]);
+                        data_f16[index] = i;
+
+                        //data_f32[index] = ggml_fp16_to_fp32(i);
+                        data_f32[index] = f16::from_bits(i).to_f32();
```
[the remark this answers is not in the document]
FloppyDisck: That function requires context to be initialized since it just uses a lookup table to find the appropriate value.

philpax:
I've updated the PR, but the output seems to be incorrect:
```
thread 'main' panicked at 'Could not load model: TensorWrongSize { tensor_name: "tok_embeddings.weight", path: "../llama-models/v0/ggml-model-q4_0-rust.bin" }', llama-cli/src/cli_args.rs:319:14
```

Probably an assumption somewhere that I broke. Need to look into it further - any ideas?

I'd also like to support loading unversioned models and GGJT, so this is going to be a bit of a headache in general :(

FloppyDisck:
Looking at your commits, I found a couple of places where it could've broken, Ill check it out and see if it fixes it.

FloppyDisck:
@philpax Regarding supporting other types of models, if you can provide the relevant issues I could research into making those work.

ggml/src/lib.rs
```diff
@@ -466,3 +466,49 @@ fn i32_to_usize(val: i32) -> usize {
 fn i64_to_usize(val: i64) -> usize {
     usize::try_from(val).unwrap()
 }
+
+/// Quantizes `src` into `dst` using `q4_0` quantization.
+///
+/// # Safety
+///
+/// You must ensure the arrays passed in are of the correct size.
```
KerfuffleV2, line 474: We should know the correct sizes here, right? I'd really suggest just adding a couple asserts to make sure it's correct, or ideally have the functions return `Result` so it can be handled more gracefully than just panicking.

Either way is better than possibly allowing memory errors though. Same thing for `quantize_q4_1`.

philpax: Yes, that's something I added as I wasn't sure when I was looking over the code: we should definitely encode these constraints more directly now that we have a CLI tool that uses them.

KerfuffleV2: Another thing I did in my fork is change the view functions to take the number of elements rather than bytes. It doesn't seem like there's an actual use to having to pull up and multiply by the element size all the time. You'd never want to offset into half of a float or something, right?

philpax:
Ok, updated to the latest `main`, haven't tested if it works. Couple of notes:

1) It would be *really* nice to support GGML and GGJT, not just GGMF. #125 should make this much easier, but could be a while off.
2) I think `f32` as the element type for `dst` of the `ggml` quantize functions is wrong. The resulting values will not be floats in any meaningful sense. `u32` is still wrong, but makes it clearer that it's a different data type.
3) It'd be a good idea to add these asserts to the `ggml` quantize functions (assuming that they're correct):
```rust
assert_eq!(src.len(), n as usize);
assert_eq!(dst.len(), n as usize);
assert!(hist.len() >= 16);
```

ggml/src/lib.rs
philpax: Further discussion is in the PR #128

KerfuffleV2:
> `u32` is still wrong, but makes it clearer that it's a different data type.

What about `u8`? When I see that, I think "just a bunch of bytes".

philpax:
> What about `u8`? When I see that, I think "just a bunch of bytes".

The issue is the size - it should be equivalent to the size of the original array in bytes. I guess we could shove 4*size onto the user, it's not that big of a deal

philpax:
Merged in main again, comments/questions from above still apply

philpax:
1. Yeah, probably. If #125 gets done sooner rather than later we can give it a shot, but if not I'm happy to merge as-is and then we can patc hit up
2. Looks like it's a `llama_buffer` in the original code https://github.com/ggerganov/llama.cpp/blob/be87b6ed20a5f7528bf491a83e759a9fc6a24fea/llama.cpp#L1596 but that's fine, we could just use a `Vec<u8>` here.

Cool then, let's see if #125 happens soon and in the meantime we can fix 2/3.

philpax:
#125 is going to be merged soon if all goes well, but its ggml-format loader doesn't work in its current stage. Given that, I think we're OK to merge this once that's in. Let's try to get support for the other formats as soon as possible, but I won't let that block merging.

philpax:
I implemented write support for the loader (now `ggml-format`), so now that `quantize` can accept any single-file GGML/GGMF/GGJT and quantize it to GGJT q4_0 or q4_1. Seems to work from my testing, but I don't have that many f16 models lying around.

PR #91, opened 2023-03-29
Added support to cached sessions in REPL mode

### LINKED ISSUE #88: Using cached sessions in REPL mode ?
Hi everyone!

I've been using  `llama-rs` recently and I especially love the caching session feature. I've been building an API that allows users to maintain multiple conversation threads and switch between them on the fly.

This means I'm currently caching each conversation using a unique id, and loading them as needed depending on which conversation the user is talking to. It works great and the response time are pretty fast on deep conversations but it also means I need to load the weights into memory each time. 

As far as I know it's not possible to use cached sessions in REPL mode. On some setups with slower disk I/O this is now the bottleneck on chats.

Would there be a way for me to achieve something like this ? To recap:
* Binary stays loaded in memory in REPL mode.
* For every prompt in REPL mode, pass a **different** cached session so it can load context before answering, and save the results after. 

Feel free to let me know if you think this is technically possible and just not exposed in the CLI, or if there's some deeper issue making this harder to achieve.

Thank you for your time, I'm a big fan of this project. :D

Metalflame12:
Closes #88

### DIFF
--- a/llama-cli/src/main.rs
+++ b/llama-cli/src/main.rs
@@ -16,13 +16,13 @@ fn repl_mode(
     vocab: &llama_rs::Vocabulary,
     params: &InferenceParameters,
     session_params: &InferenceSessionParameters,
+    mut session: InferenceSession,
 ) {
     let mut rl = rustyline::DefaultEditor::new().unwrap();
     loop {
         let readline = rl.readline(">> ");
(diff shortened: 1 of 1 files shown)


philpax:
Congratulations on your first PR 🚀

PR #94, opened 2023-03-31
Fix for the --batch-size parameter

Metalflame12:
The default value was being used regardless of the value provided by the user.


philpax:
Good spot!

PR #104, opened 2023-04-03
Fix num_threads flag being ignored

setzer22:
Some previous changes broke the num_threads flag in `llama_cli`. This PR restores the original behavior of obeying the value for the -t flag, and only using the autodetected value when none is set.


philpax (approved):
LGTM once formatting's fixed

PR #107, opened 2023-04-04
Update to latest ggml and bring back bindgen as a dev dependency

setzer22:
The latest updates to GGML had introduced changes in lots of functions (many values have changed from i32 to i64) and some structs had its layout changed, which caused UB because our bindings were not up to date.

This PR brings back bindgen, but adds it as a dev dependency so it doesn't impact user builds. What I did is add a `generate_bindings.rs` as a script, because apparently there's no way to make binaries aware of dev dependencies (only tests, benchmarks and examples).

So, after this PR, the process to update GGML is as follows:

- Update the `ggml.c` and `ggml.h` in `ggml-sys/ggml`.
- In that same folder, take note of the llama.cpp version these files were taken from  inside the CREDITS.md file.
- Run the bindgen script:
```shell
$ cd ggml-sys
$ cargo run --example generate_bindings
```
- Fix any compiler errors that pop up due to the new version of the bindings and test the changes.


philpax (approved):

philpax:
Looks good to me, but

> What I did is add a generate_bindings.rs as a script, because apparently there's no way to make binaries aware of dev dependencies (only tests, benchmarks and examples).

I'd suggest creating a separate binary (`generate-bindings` or `ggml-update-bindings`) and bring it in as a dependency there. The example thing is a hack.

> So, after this PR, the process to update GGML is as follows:

I'd suggest adding this to a CONTRIBUTING.md or similar.

PR #108, opened 2023-04-04
scripts: use GGMF version of convert-pth-to-ggml.py

philpax:
The version of the `convert-pth-to-ggml` script we had was for GGJT, which we don't support. This fixes it by pulling the script from https://github.com/ggerganov/llama.cpp/blob/a5c42c4b13b3be9e58fe8f9adbb6ee60417674a6/convert-pth-to-ggml.py


setzer22:
I didn't realize they replaced the original script... 😬 

Thanks for clearing this up! LGTM. We can merge right away to prevent further confused people

setzer22 (approved):

PR #111, opened 2023-04-05
fix: cross compile from non-apple silicon host to apple silicon

hlhr202:
Hi, this should solve the apple m1 targeting cross compile problem. We should check if host platform is same as target platform, if same, we enable mcpu=native, otherwise we manually setup the flags


philpax (approved):
This is fine for now, we can figure out a long-term solution with more architectures later

philpax:
Can you chuck a `#[allow(clippy::single_match)]` in there?

hlhr202:
> Can you chuck a `#[allow(clippy::single_match)]` in there?

done

Issue #93, opened 2023-03-30
Support the new `mmap`-able ggml format


philpax:
Justine's managed to land her mad-lass mmap-optimised format into llama.cpp. We should support loading this format in - and if we're smart about it, we should also support `mmap`ing it in. This should hopefully be easier to do in Rust than it is in C++!

philpax:
https://github.com/ggerganov/llama.cpp/pull/711#issuecomment-1495320638

Narsil:
@philpax would you be interested if I added proper support in https://github.com/huggingface/safetensors/pull/197 ?

The format allows for `mmap` (or not, but it will currently align buffers for zero-copy loads).
But it is a "pure" rust format (which is also readable from Python)

philpax:
Hm, don't see why not! It'll depend on model availability, but I imagine we'll start seeing models released in ST format.

katopz:
FYI: note for `mmap` https://justine.lol/mmap/

KerfuffleV2:
> FYI: note for mmap https://justine.lol/mmap/

Seems like some crazy misinformation to me. I've never even seen a multipart GGML file. The whole dataset is also needed for processing each token, so you can't practically use models larger than memory because it will require repeatedly loading the data from disk.

"100x faster" — maybe if the entire thing is already in the buffer cache, but that's only possible when you have enough memory available to load the whole model.

mmap is great and some overhead can be avoided (sometimes) when using it, but it's not magic.

KerfuffleV2:
> Converting any of the multipart .pths will result in multipart GGMLs.

I see. It may make a bigger difference in that case, but they also could have just changed the conversion process to make stuff contiguous without having to mess with the final file format.

> Yes, I think the primary benefit is in repeated executions so that the model remains resident across executions.

Well, it's just the OS buffer cache. The OS will cache stuff whether you're using mmap or just reading it normally. mmaping may avoid some copying data, but loading the model was already very fast when the cache was hot.

> My understanding is that it's laying out the data so that it matches the layout in memory. See https://github.com/rustformers/llama-rs/pull/114.

Nice, someone else is dealing with it! Although, right now the new format part is just a `todo!("new format here)` so that doesn't really help with understanding the change at present.

philpax:
> It may make a bigger difference in that case, but they also could have just changed the conversion process to make stuff contiguous without having to mess with the final file format.

[...]

PR #85, opened 2023-03-29
WIP: Bloom Inference

hhamud:
Completes #45 

Will refactor and remove most of the duplicate code before merging. 

* Adds a build flag for apple silicon
* Adds `ggml_alibi`, `ggml_compute_alibi_forward_alibi_f32`, `ggml_compute_forward_alibi_f16`, `ggml_compute_forward_alibi` to `ggml.c`
* Adds in `ggml_view_2d`, `ggml_alibi` and `ggml_gelu` to `lib.rs` in ggml-raw crate.
* Adds a bloom mode flag to the CLI.
* Splits up the original library into a `commons` folder and the models into a `models` folder
* Moved some code into functions in `main.rs` in `llama-cli`
* Updates ggml.h

Issue #11, opened 2023-03-15
Warning: Bad token in vocab at index xxx


CheatCod:
running `cargo run --release -- -m ~/dev/llama.cpp/models/7B/ggml-model-f16.bin -f prompt` gives a bunch of "Warning: Bad token in vocab at index..." 
![image](https://user-images.githubusercontent.com/47200772/225275909-0536024d-4103-4113-ad3e-10cef33365d4.png)

The path points to ggml converted llama model, which I have verified that they work with llama.cpp

setzer22:
Thanks for reporting @CheatCod! We are aware of the issue. Apparently, some models have tokens in the embedded vocabulary that use invalid UTF-8 codepoints. This is an error, but not an irrecoverable one. The C++ code simply ignores this (C++ strings are just byte arrays, while in Rust, they're required to hold valid UTF-8), but I thought it would be a good idea to at least warn the users.

There is some discussion about this at #3. Some of us don't see the errors, while others do. If I had to take a guess, this is something in the `convert-pth-to-ggml.py` script which, in one case is happily writing the invalid codepoints, and in other cases (like me, and others who reported the same), is substituting the invalid characters by the replacement character.

mwbryant:
Hopefully this will be less intrusive after #10 merges and we have a better logging crate.  Maybe it shouldn't even be a warn! and should be an info! (or even not print at all) because it really doesn't hurt anything according to @setzer22's digging into the actual token.

setzer22:
Yup, I'd say turning it into `info` would be nice to avoid log spam, since the issue seems to be quite common.

philpax:
I've downgraded it to `info` in #10 in the CLI (which now controls the logs).

setzer22:
Not sure what to do about this. We often get reports about this issue, so leaving it open sounds like a good idea to avoid duplicates.

But also, there's very little we can do about those non-utf8 tokens in the vocabulary, they are most certainly an error and affect lots of users.

So, what should we do? One option is to downgrade the error even further down to `trace` level. Another is to simply leave things as is, since people can use the RUST_LOG env variable to ignore all `info` messages. But people will probably still want to see the other loading messages.

philpax:
I think leaving it as-is is fine: we can hopefully figure out what's generating those invalid tokens when we get to #21.

mwbryant:
[...]

PR #114, opened 2023-04-06
Add GGJT loader

iacore:
Related to #62 #93 

single-file model format magic=ggjt

### DIFF
--- a/llama-rs/src/lib.rs
+++ b/llama-rs/src/lib.rs
@@ -60,6 +60,15 @@ struct Layer {
     w3: ggml::Tensor,
 }
 
+
+/// Model Version
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub(crate) enum ModelVersion {
+    GGMF,
+    GGJT,
+    Unversioned,
+}
+
 /// The weights for the LLaMA model. All the mutable state is split into a
 /// separate struct `InferenceSession`.
 pub struct Model {
@@ -75,6 +84,8 @@ pub struct Model {
     tensors: HashMap<String, ggml::Tensor>,
 
     mmap: Option<Mmap>,
+    
+    version: ModelVersion,
 
     // Must be kept alive for the model
     _context: ggml::Context,
@@ -604,10 +615,10 @@ impl Model {
         let mut reader = BufReader::new(&file);
 
         // Verify magic
-        let model_type: ModelType = match read_u32(&mut reader)? {
-            ggml::FILE_MAGIC_GGMF => ModelType::GGMF,
-            ggml::FILE_MAGIC_GGJT => ModelType::GGJT,
-            ggml::FILE_MAGIC_UNVERSIONED => ModelType::Unversioned,
+        let model_type: ModelVersion = match read_u32(&mut reader)? {
+            ggml::FILE_MAGIC_GGMF => ModelVersion::GGMF,
+            ggml::FILE_MAGIC_GGJT => ModelVersion::GGJT,
+            ggml::FILE_MAGIC_UNVERSIONED => ModelVersion::Unversioned,
             _ => {
                 return Err(LoadError::InvalidMagic {
                     path: main_path.to_owned(),
@@ -617,13 +628,13 @@ impl Model {
 
         // Load format version
         match model_type {
-            ModelType::GGMF | ModelType::GGJT => {
+            ModelVersion::GGMF | ModelVersion::GGJT => {
                 let _version: u32 = match read_u32(&mut reader)? {
                     ggml::FORMAT_VERSION => ggml::FORMAT_VERSION,
                     version => return Err(LoadError::InvalidFormatVersion { value: version }),
                 };
             }
-            ModelType::Unversioned => {}
+            ModelVersion::Unversioned => {}
         }
 
         // =================
@@ -660,8 +671,8 @@ impl Model {
             for i in 0..hparams.n_vocab {
                 let len = match model_type {
                     // `read_i32` maybe a typo
-                    ModelType::GGMF | ModelType::Unversioned => read_i32(&mut reader)? as usize,
-                    ModelType::GGJT => read_u32(&mut reader)? as usize,
+                    ModelVersion::GGMF | ModelVersion::Unversioned => read_i32(&mut reader)? as usize,
+                    ModelVersion::GGJT => read_u32(&mut reader)? as usize,
                 };
                 let maybe_word = if len > 0 {
                     read_string(&mut reader, len)
@@ -682,12 +693,12 @@ impl Model {
 
                 // Token score, currently unused
                 match model_type {
-                    ModelType::GGMF | ModelType::GGJT => {
+                    ModelVersion::GGMF | ModelVersion::GGJT => {
                         if let Ok(score) = read_f32(&mut reader) {
                             id_to_token_score.push(score);
                         }
                     }
-                    ModelType::Unversioned => {
+                    ModelVersion::Unversioned => {
                         // Legacy model, set empty score
                         id_to_token_score.push(0.);
                     }
@@ -815,11 +826,12 @@ impl Model {
                 tensors,
                 _context: context,
                 mmap: None,
+                version: model_type,
             }
         };
 
         match model_type {
-            ModelType::GGMF | ModelType::Unversioned => {
+            ModelVersion::GGMF | ModelVersion::Unversioned => {
                 let file_offset = reader.stream_position()?;
                 drop(reader);
                 load_weights_ggmf_or_unversioned(
@@ -829,7 +841,7 @@ impl Model {
                     &model,
                 )?
             }
-            ModelType::GGJT => {
+            ModelVersion::GGJT => {
                 let mmap = unsafe { Mmap::map(&file)? };
                 load_weights_ggjt(
                     &mut reader,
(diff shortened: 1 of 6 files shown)


philpax:
Very nice! Glad someone took this on - we were discussing this on the Discord but decided to wait until upstream figured out what they wanted to do.

There's overlap with #85, so the merge in the future might be tricky. Just a heads-up.

iacore:
I ported the code from C++.

I'm sure the reading part is correct. Maybe Tensor setup has changed? Or ggml has also changed?

philpax:
I'd take the same model in ggmj and ggjt format and compare the loaded tensors. I'm guessing it's probably a misalignment or something similar.

iacore:
Is it normal to warn for bad tokens? The bad tokens are single-byte string >128. That's invalid UTF-8, but the tokenizer don't care?

philpax:
Yes, we're discussing that over at #11. The short explanation is that the invalid tokens are not valid UTF8 by themselves, but are composed to form valid UTF8. We're figuring out what the actual solution should be.

KerfuffleV2:
Does it work with the old formats but just gets garbage when using a GGJT model or just always garbage?

iacore:
I haven't change the behavior of old formats. only GGJT produce garbage

KerfuffleV2:
## `ggml`

Fine.

## `ggmf`

```plaintext
thread 'main' panicked at 'Could not load model: TensorWrongSize { tensor_name: "tok_embeddings.weight", path: "blah.bin" }', llama-cli/src/main.rs:206:6
```

(I don't know if it's actually supposed to work here or not.)

## `ggjt`

Garbage.

KerfuffleV2:
Was that last force push just rebasing on the current version or does it involve changes to the loading that may fix stuff that previously didn't work?

iacore:
> Was that last force push just rebasing on the current version or does it involve changes to the loading that may fix stuff that previously didn't work?

just rebase

KerfuffleV2:
I think this one is GGMF: https://huggingface.co/Sosaka/Alpaca-native-4bit-ggml/

(Don't know if it's a problem to post something like that here, if so let me know and I'll edit it out after iacore sees it.)

iacore:
+1

KerfuffleV2:
Sorry, my mistake. Unfortunately, I don't really have a reasonable way to share huge files.

I've been trying to figure out what the issue with GGJT is. One thing I can say is I don't think your logic for finding the tensors/lengths/offsets has a problem.

I added some `printf`s to `llama.cpp` and the corresponding ones to `llama-rs`.

```plaintext
Load: tok_embeddings.weight, offset=432672, size=102403200
Load: norm.weight, offset=102835904, size=20480
Load: output.weight, offset=102856448, size=102403200
[...]
Load: layers.39.feed_forward.w2.weight, offset=8048282880, size=44236800
Load: layers.39.feed_forward.w3.weight, offset=8092519744, size=44236800
Load: layers.39.ffn_norm.weight, offset=8136756608, size=20480
```

Absolutely no difference in output between the C and Rust versions.

iacore:
still no clue

KerfuffleV2:
Okay, so I got it actually running inference on a GGJT model. However, what I had to do makes the mmap part pointless.

I believe the problem has something to do with how you can set a context parameters to `no_alloc`. In the llama.cpp change that added the format, they set `no_alloc` to true for the main context and then reduce the context size a lot so that GGML doesn't allocate memory for the actual tensors.

However, we're still doing the old context size calculation. I tried making the `Context::init` function take a bool for no_alloc and set it, but just got a segfault immediately.

Anyway, in `loader.rs:load_weights_ggjt` just change:

```rust
tensor.set_data(ptr as *mut std::ffi::c_void);
```

to

```rust
ptr.copy_to_nonoverlapping(tensor.data() as *mut u8, tensor.nbytes());
```

With that, it runs just fine on the GGJT model. Loading speed seems normal compared to the current version.

Obviously it's silly to `mmap` and just copy the data immediately. What I'd recommend is just ditching `mmap` right now and simply reading into the tensor data instead.

Then later on it will be possible to add `mmap` support as a separate thing which could work in a general way for the other formats too.

By the way, you probably need to run clippy on your changes. It's very unhappy right now!

philpax:
Yeah, I'd be happy with not supporting `mmap` right now. We can figure out what that's meant to look like once we have support for all the model types working.

KerfuffleV2:
The approach I'd go for if I was writing it would be to just have a general type that can describe the types of tensor and where they are, dimensions, etc. Something similar to this: https://github.com/KerfuffleV2/smolrsrwkv/blob/182cd3205b7a7c95571a09bcfbb954b0041e4f90/smolrwkv/src/loader.rs#L15

Then the specific file format code can just scan through the file for metadata, build a structure with a list of those things or whatever. Then there can be generic loader code that just loads tensors based on that: it could use reading files, mmap, whatever.

I think that approach would make dealing with different file formats _much_ easier.

philpax:
I'd be OK with that. It'd also help with #117 / #84.

iacore:
ship it

KerfuffleV2:
@iacore I was experimenting with trying to clean it up also. It does seem like reading is way slower than the copying from `mmap` approach. I don't know why.

My change looks like this:

```rust
        let offset_curr = reader.stream_position()?;
        let offset_aligned: u64 = (offset_curr + 31) & !31;

        reader.seek_relative((offset_aligned - offset_curr) as i64)?;
        let td =
            unsafe { std::slice::from_raw_parts_mut(tensor.data() as *mut u8, tensor.nbytes()) };

        reader.read_exact(td)?;
        total_loaded_bytes += tensor.nbytes();
```

So, uhh... I guess maybe just keep mmap?

iacore:
for me, seq. read is faster than mmap

KerfuffleV2:
Are you saying you tried the `copy_to_nonoverlapping` version and it was slower than the current version that changed to `BufReader`?

iacore:
> Are you saying you tried the `copy_to_nonoverlapping` version and it was slower than the current version that changed to `BufReader`?

yes

KerfuffleV2:
Ahh, why can't life be simple for once?

What OS are you using, out of curiosity?

iacore:
Linux.

I think this branch is done. I'll not touch it again.

KerfuffleV2:
I'm on Linux as well.

Sorry for the confusion. I've been trying both versions and I don't get consistent results. No sure what's going on, but I don't think there's a problem with the current approach.

> I'll not touch it again.

How can you say that when Clippy is still sad?

You can probably just nuke the `set_data` method, it doesn't seem clear that there's even a way to successfully use it at the moment.

iacore:
> You can probably just nuke the set_data method, it doesn't seem clear that there's even a way to successfully use it at the moment.

it probably will be useful in the future?

KerfuffleV2:
> it probably will be useful in the future?

We'd have to figure out how to actually use it in the future though. I don't think it currently can work at all, since you aren't even able to turn of `need_alloc` when creating a context so memory for tensors will always get allocated no matter what.

You'd think it would still be possible to set the tensor to point at a different chunk of memory but that didn't actually work: otherwise your first approach would have had no issues.

So my line of thinking is, it's a only a couple of lines of code to wrap a `ggml` function so it wouldn't be hard to add back in later on if it was actually needed/could be used but it's currently non-functional so it may as well be removed.

Just to be clear, this is just the opinion of some other random person on the internet. So take that for it's worth, I have no authority.

jon-chuang:
> I think this branch is done. I'll not touch it again.

So, we're not gonna get the new mmap format cuz poor perf?

jon-chuang:
I also got this working on new ggjt file format.

jon-chuang:
Btw, from https://justine.lol/mmap/

> Remember that progress bar which made you wait for weights to load each time you ran the command? We got rid of that. Linux users should expect a 100x improvement in load time. Windows and MacOS users should expect a 10x improvement. What this means is that tokens will start being produced effectively instantaneously when you run LLaMA, almost providing a similar UX to ChatGPT on the shell. It's important to note these improvements are due to an amortized cost. The first time you load a model after rebooting your computer, it's still going to go slow, because it has to load the weights from disk. However each time it's loaded afterwards, it should be fast (at least until memory pressure causes your file cache to be evicted).

The speedup is meant to be for subsequent loads. Did you guys check that?

jon-chuang:
Another thing to check: multiple processes utilizing the same mmaped file:
> More Processes
You can now run multiple LLaMA processes simultaneously on your computer. Here's a [video of Georgi having a conversation with four chatbots](https://twitter.com/ggerganov/status/1642115206544343040) powered by four independent llama.cpp processes running on the same Mac. So llama.cpp is not only going to be a better friend to you, it can also serve as your artificial circle of friends too. The trick that makes it possible is mmap() lets us map the read-only weights using MAP_SHARED, which is the same technique that's traditionally been used for loading executable software. So we figured, why aren't we using it to load neural network software too? Now we can.

llama-rs/src/lib.rs
```diff
@@ -828,243 +841,28 @@ impl Model {
...
+                drop(reader);
+                load_weights_ggmf_or_unversioned(
+                    file_offset,
+                    main_path,
+                    load_progress_callback,
+                    &model,
+                )?
+            }
+            ModelVersion::GGJT => {
+                let mmap = unsafe { Mmap::map(&file)? };
```
jon-chuang, line 861: Can add advise
https://docs.rs/memmap2/latest/memmap2/struct.Mmap.html#method.advise
here with
https://docs.rs/memmap2/latest/memmap2/enum.Advice.html#variant.Sequential

iacore:
Superceded by #125

Issue #115, opened 2023-04-06
Runs out of ggml context's memory pool on larger models


juho-p:
Running LLaMA 30B and 65B models:

`ggml_new_tensor_impl: not enough space in the context's memory pool (needed 1073742848, available 1073741824)`

followed by Segmentation Fault.

To reproduce, run following: (sorry for silly prompt (mostly taken from llama.cpp examples), it's almost what my current project "poop-gpt" uses.)

```
cargo run --release --bin llama-cli -- infer -m ../models/LLaMA/30B/ggml-model-q4_0.bin --num-ctx-tokens 1024 --repeat-last-n 128 --seed 1 -p "Text transcript of a never ending dialog, where User interacts with an AI assistant named Nakki.
Nakki is helpful, kind, honest, good at writing and never fails to answer User’s requests immediately and with precision.
There are no annotations like (30 seconds passed...) or (to himself), just what User and Nakki say alound to each other.
The dialog lasts for years, the entirety of it is shared below. It's 10000 pages long.
The transcript only includes text, it does not include markup like HTML and Markdown.
User: Hello, Nakki!
Nakki: Hello User! How may I help you today?
User: What time is it?
Nakki: It is 12:12.
User: What year is it?
Nakki: We are in 2023.
User: What is a cat?
Nakki: A cat is a domestic species of small carnivorous mammal. It is the only domesticated species in the family Felidae.
User: How many legs?
Nakki: A cat has four legs.
User: How do I pass command line arguments to a Node.js program?
Nakki: The arguments are stored in process.argv.
    argv[0] is the path to the Node. js executable.
    argv[1] is the path to the script file.
    argv[2] is the first argument passed to the script.
    argv[3] is the second argument passed to the script and so on.
User: How do I loop a list in Python?
Nakki: To loop a list in Python, you can use for-statement
for item in list:
    print(item)
User: Name a color.
Nakki: Blue
User: How to configure git pull to allow non fast forwarding merges?
"
```

Issue #77, opened 2023-03-26
Swap strategy for infinite output


setzer22:
As discussed in https://github.com/ggerganov/llama.cpp/issues/71#issuecomment-1483907574

The idea is to achieve a naive implementation for infinite output generation using a strategy that simply clears the context window (you can keep the original prompt around), and simply starts adding new tokens.

This is a hack that doesn't properly leverage the advantages of the attention mechanism: When the context window gets full, the transformer's hidden state has information about more than just the last 2048 tokens, because this information is there indirectly embedded in the outputs for the self-attention mechanism. For example, if token 25 attended to tokens 10 and 12, even when tokens 10 and 12 fall outside the context window, a lot of information about these tokens will still be encoded at position 25.

A solution that slides the context window would achieve a gradually "fading" context window, instead of something where the transformer 100% forgets about a word the moment it's fallen out of the context window. I have some reason to suspect systems like ChatGPT are relying on a mechanism like this based on their ability to consistently recall parts of the conversation that occured way before the token window was exceeded.

By doing the swap trick proposed here, the transformer will effectively forget all prior context whenever the swap occurs, and there will be a lag spike due to the last few tokens having to be reprocessed. So this is very much non-ideal. However, since llama.cpp has recently implemented this, I feel like we should at least add this version too until someone can figure out a real solution.

jon-chuang:
Yes, llama.cpp implements a "hacky" method like this, it takes the last $k$ tokens + first $n$ "prompt" tokens when the window becomes full

olexiyb:
There is a pull request to solve this, please review
https://github.com/rustformers/llm/pull/424

PR #116, opened 2023-04-06
Reserve more eval memory and use ggml scratch buffers

juho-p:
This attempts to fix https://github.com/rustformers/llama-rs/issues/115

Not the brightest of code. Scratch buffer usage is imitated from llama.cpp. `evaluate` context memory reservation is also imitated there (or at least that's the attempt). Note that I didn't actually go through all the things in llama.cpp, only the ones related to context memory size, so I might be missing something.

I'm quite sure this allocates a lot more extra memory for 7B and 13B models as well, even though I never had any issues with them running out of context memory. Scratch buffers take 1GB, and that's for every `InferenceSession`.

Maybe it would make sense to only use scratch buffers for inferring bigger models only? Though I think llama.cpp uses them always. Also, if there are multiple sessions, scratch memory buffers are not shared between them, but they could (just make sure only run one `evaluate` at a time somehow).

Anyway, with these changes I couldn't manage to run out of context memory any more, even with 65B model.

### DIFF
--- a/ggml/src/lib.rs
+++ b/ggml/src/lib.rs
@@ -379,8 +379,10 @@ impl Context {
         unsafe { ggml_sys::ggml_used_mem(self.ptr.as_ptr()) }
     }
 
-    /// Set scratch buffer
-    pub fn use_scratch(&self, scratch_buffer: Option<&mut Buffer>) {
+    /// Sets the scratch buffer to be used by this [Context].
+    ///
+    /// If `scratch_buffer` is `None`, the scratch buffer will be disabled.
+    pub fn use_scratch<'a>(&'a self, scratch_buffer: Option<&'a mut Buffer>) {
         let (size, data) = if let Some(buffer) = scratch_buffer {
             (buffer.data.len(), buffer.data.as_ptr() as *mut c_void)
         } else {
@@ -410,22 +412,28 @@ impl Drop for Context {
     }
 }
 
-/// Pre-allocated buffer
+/// A buffer of memory that can be used as a scratch buffer for a [Context].
+///
+/// See [Context::use_scratch].
 pub struct Buffer {
-    data: Vec<u8>,
+    data: Box<[u8]>,
 }
 
 impl Buffer {
-    /// Creates new buffer
+    /// Creates a new buffer of the specified size.
     pub fn new(size: usize) -> Self {
         let mut data: Vec<u8> = Vec::with_capacity(size);
-        // SAFETY: contents are left uninitialized. Don't use them.
+
+        // SAFETY: The contents are intentionally uninitialized, as they will be passed to
+        // the ggml C API which will fill them with data.
         #[allow(clippy::uninit_vec)]
         unsafe {
-            data.set_len(size)
-        };
+            data.set_len(size);
+        }
 
-        Buffer { data }
+        Buffer {
+            data: data.into_boxed_slice(),
+        }
     }
 }
 
--- a/ggml/src/lib.rs
+++ b/ggml/src/lib.rs
@@ -303,6 +303,26 @@ impl Context {
     pub fn used_mem(&self) -> usize {
         unsafe { ggml_sys::ggml_used_mem(self.ptr.as_ptr()) }
     }
+
+    /// Set scratch buffer
+    pub fn use_scratch(&self, scratch_buffer: Option<&mut Buffer>) {
+        let (size, data) = if let Some(buffer) = scratch_buffer {
+            (buffer.data.len(), buffer.data.as_ptr() as *mut c_void)
+        } else {
+            (0, std::ptr::null_mut())
+        };
+        // SAFETY: this just passes (most likely uninitialized) memory buffer to the ggml C API
+        unsafe {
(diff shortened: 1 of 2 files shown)


ggml/src/lib.rs
```diff
@@ -303,6 +303,23 @@ impl Context {
     pub fn used_mem(&self) -> usize {
         unsafe { ggml_sys::ggml_used_mem(self.ptr.as_ptr()) }
     }
+
+    /// Set scratch buffer
+    pub fn use_scratch(&self, scratch_buffer: Option<&Buffer>) {
```
philpax, line 308: Shouldn't this be a `Option<&mut Buffer>`?

philpax:
man I do not enjoy the amount of magic constants we're accumulating, but aside from the `&mut` thing I think this is fine. Does `llama.cpp` always reserve 1GB of scratch, even for 7B?

llama-rs/src/lib.rs
```diff
@@ -103,6 +105,9 @@ pub struct InferenceSession {
 
     /// The logits that were last predicted by the network. Zeroed out otherwise.
     last_logits: Vec<f32>,
+
+    /// Scratch buffers
+    scratch: [ggml::Buffer; 2]
```
philpax, line 110: Do we know why it needs two buffers?

ggml/src/lib.rs
iacore: the infered lifetime is wrong. The buffer is consumed.

`&mut Buffer` should at least live until `&self` is gone

iacore: Signature could be

`pub fn use_scratch(&self, scratch_buffer: Option<&mut [u8]>)`

then `Buffer` isn't needed

iacore: I have a feeling Rust's aliasing rules would not like this.

The context should have a field like
`_scratch: Option<PhantomData<&mut [u8]>>`

llama-rs/src/lib.rs
iacore: > Do we know why it needs two buffers?

It's from llama.cpp. Same with the magic buffer sizes.

juho-p: Indeed. I don't know. And I'd _guess_ 7B wouldn't need as much, but I don't know and I didn't try, and even if I tried, I couldn't know what's the actual correct amount.

https://github.com/ggerganov/llama.cpp/blob/62cfc54f77e519057110265b52b0d614fa363e2a/llama.cpp#L63

ggml/src/lib.rs
juho-p: Yes, definitely should be mut. I'll fix it.

And yes, it could be `Option<&mut [u8]>` but I don't think there's a safe way to create uninitialized `[u8; BUF_SIZE]`, so the `Buffer` helper at least simplifies that bit slightly. Or alternatively could just create `Vec<u8>` with wanted amount of `0`s, but that does of course cause "unneeded" runtime cost.

llama-rs/src/lib.rs
juho-p: As for why it's two buffers, it's quite unclear to me. I'd guess ggml expects certain objects to live in scratch 2 while pointing to the scratch 1. I can say for sure it crashes if there's only one scratch buffer used (replace all `use_scratch` calls with the same buffer).

ggml/src/lib.rs
juho-p: Actually I also wonder if I should just make this function unsafe as well... It does crash to segmentation fault if it's used incorrecly. On the other hand I guess quite a lot of other functions end up being unsafe as well if that would be the reasoning.

llama-rs/src/lib.rs
juho-p: And for the evaluate context memory requirements:
- https://github.com/ggerganov/llama.cpp/blob/62cfc54f77e519057110265b52b0d614fa363e2a/llama.cpp#L88
- https://github.com/ggerganov/llama.cpp/blob/62cfc54f77e519057110265b52b0d614fa363e2a/llama.cpp#L443

It's not really the greatest engineering to just copy the logic from llama.cpp without understanding it fully. At the moment I don't have better ideas though. I can try to think cleaner ways to do this context memory handling maybe next week, but that could be another PR as far as I'm concerned.

ggml/src/lib.rs
philpax: It being unsafe seems reasonable to me - if the caller has to make guarantees that aren't handled by Rust in order to use it safely, it should be `unsafe`

llama-rs/src/lib.rs
philpax: Hm, OK, fair enough. We should document that we don't actually know the reasoning behind this, so it's clear that it's cargo-culting instead of some principled engineering choice.

jon-chuang:
Will this help `Context window full, stopping inference.`?

philpax:
> Will this help `Context window full, stopping inference.`?

No, that's because the model has a fixed context limit of 2048 bytes. We're investigating ways to improve this in #77.

PR #119, opened 2023-04-07
Sync to llama.cpp + GGML version as of 20230407 09:57 AM UTC.

### LINKED ISSUE #67: Making results independent from threadcount/batch size (from llama.cpp)

### LINKED ISSUE #118: Update to the latest `llama.cpp` `ggml`
Another day, another suite of changes. Since our last update, M1 and AVX inference has improved, as well as a bunch of other improvements: https://github.com/ggerganov/llama.cpp/compare/437e77855a54e69c86fe03bc501f63d9a3fddb0e..HEAD

If you're interested in tackling this, check out our contributing document: https://github.com/rustformers/llama-rs/blob/main/CONTRIBUTING.md

KerfuffleV2:
This would have been a pretty interesting first issue.

This also includes nuking the `increased_determinism` option since it's no longer needed.

I confirmed that it's possible to change the thread/batch sizes with deterministic results. I haven't don't any extensive performance testing but there didn't seem to be an obvious issue.

Other people should probably do some testing before this gets merged as it's a fairly complicated change. I can't say I really understood what I was porting across.

Closes #118
Closes #67

### DIFF
--- a/ggml/src/lib.rs
+++ b/ggml/src/lib.rs
@@ -147,6 +147,12 @@ impl Context {
         self.new_tensor_raw(raw)
     }
 
+    /// Unknown, aside from the obvious. It's transposing something!
+    pub fn op_transpose(&self, a: &Tensor) -> Tensor {
+        let tensor = unsafe { ggml_sys::ggml_transpose(self.ptr.as_ptr(), a.ptr.as_ptr()) };
+        self.new_tensor_raw(tensor)
+    }
+
     /// Unknown.
     pub fn op_get_rows(&self, a: &Tensor, b: &Tensor) -> Tensor {
         let tensor =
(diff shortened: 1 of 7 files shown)


ggml/src/lib.rs
```diff
@@ -147,6 +147,12 @@ impl Context {
         self.new_tensor_raw(raw)
     }
 
+    /// Unknown, aside from the obvious. It's transposing something!
```
philpax, line 150: love that documentation from upstream

philpax:
Seems pretty reasonable from what I can see, will need to test but have no objections to merging otherwise

ggml/src/lib.rs
KerfuffleV2: It's not from upstream, it's from me. :)

The original comment was `// ggml_transpose`

philpax: Yeah that's what I mean - upstream's documentation is... lacking 😅

KerfuffleV2: Oh, derp. My brain is slightly fried at the moment.

philpax:
Can you also update https://github.com/rustformers/llama-rs/blob/main/ggml-sys/ggml/CREDITS.txt ?

KerfuffleV2:
Can do! By the way, the contributing thing refers to `CREDITS.md`, I didn't know what it meant until now.

philpax:
> Can do! By the way, the contributing thing refers to `CREDITS.md`, I didn't know what it meant until now.

made it clearer, ty

philpax (approved):
Tested on macOS M1 and Windows x86-64 with no issues. Let's do it 🚀

PR #120, opened 2023-04-07
Update ggml.c/h to the version used in llama.cpp

iacore:


philpax:
Seems like Kerfuffle already did this in #119 ?

philpax:
Apologies for the work double-up, but #119 seems to include everything from this PR

PR #122, opened 2023-04-07
fix(llama): buffer tokens until valid UTF-8

philpax:
As discussed on Discord and in #11.

This switches the internal representation of tokens over to raw bytes, and buffers tokens until they form valid UTF-8 in `inference_with_prompt`.

Open questions:
1) Should we use `smallvec` or similar for tokens? We're going to be making a *lot* of unnecessary tiny allocations as-is.
2) `FnMut` as a bound is OK, right?

### DIFF
--- a/llama-rs/src/convert.rs
+++ b/llama-rs/src/convert.rs
@@ -49,11 +49,12 @@ fn load_vocabulary(path: &Path) -> Vocabulary {
     let mut token_to_id = HashMap::new();
     let mut max_token_length = 0;
 
+    // TODO: Does the original model use valid UTF-8 for its tokens? This seems a little suspect to me.
     for (idx, piece) in proto.get_pieces().iter().enumerate() {
-        let word = piece.get_piece().to_string();
+        let word = piece.get_piece().as_bytes();
         max_token_length = max_token_length.max(word.len());
-        id_to_token.push(word.clone());
-        token_to_id.insert(word, idx as i32);
+        id_to_token.push(word.to_owned());
+        token_to_id.insert(word.to_owned(), idx as i32);
         id_to_token_score.push(piece.get_score());
     }
     Vocabulary {
@@ -128,13 +129,20 @@ fn write_header(fout: &mut File, hparams: &Hyperparameters) -> Result<(), String
 fn write_tokens(file: &mut File, vocab: &Vocabulary) -> Result<(), String> {
     let mut values: Vec<u8> = vec![];
     for (i, token) in vocab.id_to_token.iter().enumerate() {
-        let text = match token {
-            _ if token.contains("<unk>") => " \u{2047} ".as_bytes().to_vec(),
-            _ if token.contains("s>") => vec![],
-            _ if token.len() == 6 && token.contains("<0x") => {
-                vec![u8::from_str_radix(&token[3..5], 16).unwrap()]
+        // TODO: Not sure what the behaviour should be if the token is not valid UTF-8.
+        //
+        // Switching to the HF tokenizer should fix this.
+        let text = if let Ok(token) = std::str::from_utf8(token) {
+            match token {
+                _ if token.contains("<unk>") => " \u{2047} ".as_bytes().to_vec(),
+                _ if token.contains("s>") => vec![],
+                _ if token.len() == 6 && token.contains("<0x") => {
+                    vec![u8::from_str_radix(&token[3..5], 16).unwrap()]
+                }
+                _ => token.replace('\u{2581}', " ").as_bytes().to_vec(),
             }
-            _ => token.replace('\u{2581}', " ").as_bytes().to_vec(),
+        } else {
+            token.clone()
         };
         values.extend((text.len() as i32).to_le_bytes());
         values.extend(&text);
--- a/llama-rs/src/convert.rs
+++ b/llama-rs/src/convert.rs
@@ -129,9 +129,6 @@ fn write_header(fout: &mut File, hparams: &Hyperparameters) -> Result<(), String
 fn write_tokens(file: &mut File, vocab: &Vocabulary) -> Result<(), String> {
     let mut values: Vec<u8> = vec![];
     for (i, token) in vocab.id_to_token.iter().enumerate() {
-        // TODO: Not sure what the behaviour should be if the token is not valid UTF-8.
-        //
-        // Switching to the HF tokenizer should fix this.
         let text = if let Ok(token) = std::str::from_utf8(token) {
             match token {
                 _ if token.contains("<unk>") => " \u{2047} ".as_bytes().to_vec(),
--- a/llama-rs/src/lib.rs
+++ b/llama-rs/src/lib.rs
@@ -570,7 +570,7 @@ impl Model {
     pub fn load(
         path: impl AsRef<Path>,
         n_context_tokens: usize,
-        load_progress_callback: impl Fn(LoadProgress),
+        mut load_progress_callback: impl FnMut(LoadProgress),
     ) -> Result<(Model, Vocabulary), LoadError> {
         use std::fs::File;
         use std::io::BufReader;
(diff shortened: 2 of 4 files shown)


KerfuffleV2:
Nice, this seems to fix my problems. I can't test with Vicuna because it's GGJT.

***

**我会讲你一个关于狐狸的故事：**Once upon a time, there was an enchanted village of Kitsune. The villagers were very proud and pleased to live in such a beautiful place with bounteous natural resources. However, they soon realized that the village faced one great danger - the dragon who lived deep beneath them, guarding all sorts of magical treasures it had accumulated over its long life span... 

***

The prompt is the bold part. It also seems to work fine with normal text, and I tested it against the main branch with a seed and got the same output in both cases (not a very extensive test).

I wouldn't _really_ worry about performance too much for this since who's generating more than 10 tokens a second and the context limit is 2048, so... It's going to be pretty insignificant in terms of effects.

If you actually cared about allocations, probably the best way would be to just preserve the buffer. You can have the callback pass in a mutable reference to copy the completed token into when it's ready. That way both buffers only need to get allocated once and just live for the the length of the session or whatever.

philpax:
Eh, I'm not so worried about the allocations as much as I am with cache coherency. We'd be allocating lots of tiny little buffers that could just as well be inline. 

You might be right though, we can figure that out later.

iacore:
Some with Vicuna

```
>> hello
⣽ 
 こんにちは、GPT-4についての������やご������ください。よろしくお���いします！—これと���った���みで���望的な状���でしたが、何かあり得るのはGPT-4について教えてほしい������やご
### Human: もう少す���ません。お���いします！—これと���った���みで���望的な状���でしたが、何かあり^C⏎
```

:facepalm: the model isn't trained to speak Unicode codepoints coherently

KerfuffleV2:
Are you saying it's worse than it was originally? The version I tried with at least seemed to do a reasonable just with Mandarin, not sure about Japanese.

As far as I know they really were only trained on English so it's not surprising if their non-English output is less than ideal.

iacore:
> Are you saying it's worse than it was originally? The version I tried with at least seemed to do a reasonable just with Mandarin, not sure about Japanese.

I haven't tried this patch yet.

> As far as I know they really were only trained on English so it's not surprising if their non-English output is less than ideal.

Not the point. The model was definitely awarded for partial codepoint during training.

iacore:
I found a fix.

I set the logits of invalid tokens to 0.0.

Here's Vicuna speaking fluently.

```
>> こんにちは
⢿ 、

APIのテストを行うための`requests`と`json`クライアントを作成します。そして、Win32网关（win32net.exe）を使用してWindowsファイアウォール上にプロキシサーバーIP地址へのリンックを取得します。これは、APIが起動中の状态である場合や、コマンドレザスのタイムアウトなどから発生可能とさ
### Human: ごめんに、日本^C⏎      
```

iacore:
Better fix:
```
❯ git diff llama-rs/
llama-rs/src/lib.rs --- Text (2 errors, exceeded DFT_PARSE_ERROR_LIMIT)
1386 1386         {
1387 1387             let scale = 1.0 / params.temperature;
1388 1388             for (i, &logit) in logits.iter().enumerate() {
.... 1389                 if (131..=258).contains(&i) {
.... 1390                     continue;
.... 1391                 };
1389 1392                 let tid = i as TokenId;
1390 1393 
1391 1394                 let val = if let Some(logit_override) = params.bias_tokens.get(tid) {
```

KerfuffleV2:
> I found a fix. I set the logits of invalid tokens to 0.0.

Which ones? The token might be invalid individually, but get combined with other tokens to form a valid unicode character. So if you just set them all to 0.0, you'll prevent it from expressing any unicode characters that aren't valid individually.

iacore:
> Which ones?

All of them.

> prevent it from expressing any unicode characters where the components aren't all valid individually.

I think it's "unicode codepoints not present in the vocabulary as a standalone token".

KerfuffleV2:
> I think it's "unicode codepoints not present in the vocabulary as a standalone token".

Right, but LLMs can combine those tokens that can't stand alone to create ones that can. If you remove all the ones that are invalid individually, that will limit the LLM's ability to express certain things. For example, it may not be able to use emoji (unless the emoji exists as a complete token in its vocabulary already).

iacore:
llama-rs can't tokenize this yet (using Vicuna)

```
### Human: This is the emoji for "cat": 🐱. What is the emoji for "dog"?
### Assistant: 
```


```
[2023-04-07T21:20:45Z ERROR llama_cli] Failed to tokenize initial prompt.
```

KerfuffleV2:
> llama-rs can't tokenize this yet (using Vicuna)

That's what this pull is intending to fix. Or do you mean it doesn't work even with this pull?

iacore:
> That's what this pull is intending to fix. Or do you mean it doesn't work even with this pull?

Sorry. This patch works for me.

llama-rs/src/convert.rs
```diff
@@ -128,13 +129,20 @@ fn write_header(fout: &mut File, hparams: &Hyperparameters) -> Result<(), String
...
     let mut values: Vec<u8> = vec![];
     for (i, token) in vocab.id_to_token.iter().enumerate() {
-        let text = match token {
-            _ if token.contains("<unk>") => " \u{2047} ".as_bytes().to_vec(),
-            _ if token.contains("s>") => vec![],
-            _ if token.len() == 6 && token.contains("<0x") => {
-                vec![u8::from_str_radix(&token[3..5], 16).unwrap()]
+        // TODO: Not sure what the behaviour should be if the token is not valid UTF-8.
+        //
+        // Switching to the HF tokenizer should fix this.
```
setzer22, line 134: There shouldn't be any non-utf8 tokens here. Any non-utf8 is already represented through the byte fallback (i.e. the `<0x...>`) tokens, which we handle here.

I've seen two kinds of behaviors from llama.cpp here, one is to write the unicode replacement character �. The other is simply to dump invalid UTF-8 into the file. That said, this was in relation to byte fallback tokens and has since been fixed, so I'm not aware of any other situation where invalid UTF-8 would be stored in the vocabulary other tha a corrupt / malformed model file.

I think a warning here would be justified :thinking:

llama-rs/src/lib.rs
```diff
@@ -386,14 +386,6 @@ impl std::fmt::Display for TokenBias {
...
     /// The hyperparameters have been loaded from the model.
     HyperparametersLoaded(&'a Hyperparameters),
-    /// A bad token was encountered during the loading process.
-    ///
-    /// This can be ignored, but invalid tokens will be replaced with
-    /// the `�` character.
-    BadToken {
-        /// The index within the vocabulary.
-        index: usize,
-    },
```
setzer22, line 396: So glad we finally get rid of this :smile:

llama-rs/src/lib.rs
```diff
@@ -1505,15 +1499,19 @@ impl InferenceSession {
         prompt: &str,
         maximum_token_count: Option<usize>,
         rng: &mut impl rand::Rng,
-        callback: impl Fn(&str) -> Result<(), E>,
+        mut callback: impl FnMut(&str) -> Result<(), E>,
```
setzer22, line 1502: :+1:  Nice catch. It's always good API design to make callbacks FnMut when we don't have special immutability requirements on our end

llama-rs/src/lib.rs
```diff
@@ -1738,3 +1745,41 @@ impl Vocabulary {
...
+#[derive(Clone, PartialEq, Default)]
+pub struct TokenUtf8Buffer(Vec<u8>);
+impl TokenUtf8Buffer {
+    /// Create a new buffer.
+    pub const fn new() -> Self {
+        Self(vec![])
+    }
+
+    /// Add a token to the buffer. If the buffer contains a valid string of UTF-8 text,
+    /// it is returned and the buffer is cleared for next use.
```
setzer22, line 1762: One thing I'm not sure about this design is what would happen if invalid UTF-8 ends up being in the stream somehow. After all, the LLM is not 100% reliable in that regard, so if it uses the byte fallback tokens and produces an invalid sequence of bytes, `str::from_utf8` will fail from there onwards, and there doesn't seem to be a way to get out of that situation. From the user end, no more tokens would be produced.

I think one way to solve this would be to attempt an error recovery strategy that skips any error bytes until it finds a substring that is valid utf-8. That is, try parsing the slice after dropping the first i tokens, for an i ranging from 0 to `&self.0.len() - 2`. If any `i` parses correctly, any previous bytes should be printed as replacement characters (�) instead. This has some "accidentally quadratic" potential, but I think in practice the sequences will be too short for this to matter.

I'm not an expert in how UTF-8 works, but I think this behaviour would be correct, because in UTF-8 no character is a prefix of any other character? But I'm not 100% sure on this.

setzer22 (changes requested):
This looks great! :smile:  Only one big comment w.r.t. error recovery but other than that we should be good to go.

PR #123, opened 2023-04-07
refactor(llama): remove bincode

philpax:
While writing up #122, I realised it doesn't really make sense to be prescriptive about how to read or write a snapshot to disk. This PR moves all read/write logic to the CLI, and leaves the snapshot/ref as Serde-compatible so that users can make their own decisions on how to snapshot.

The main benefit of this is that it removes the `bincode` dependency from the library.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -460,6 +460,7 @@ checksum = "f051f77a7c8e6957c0696eac88f26b0117e54f52d3fc682ab19397a8812846a4"
 name = "llama-cli"
 version = "0.1.0"
 dependencies = [
+ "bincode",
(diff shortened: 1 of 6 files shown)


setzer22 (approved):
LGTM! :+1:  Not much to add, the change makes sense, we don't need bincode and any `llama-rs` user can pick the serialization library that best fits them.

But I think it would be worth it to add a comment on `InferenceSession` to make it clear to users to always pick a binary format (i.e. no JSON or RON) because serializing large number matrices in this format could easily increase the size (of the already quite large snapshots) by several orders of magnitude.

PR #125, opened 2023-04-08
Standalone loader

iacore:
I separated the loading logic into a crate.

supercedes  #114  #122


philpax:
Merged in the latest `main` and confirmed that it works with 7B GGML and 13GB GGJT. 

Do you have an ETA for when you'll be done with the separate crate?

Also, do you think it would be possible to restructure the loader so that it can parse any valid GGML file and do LLaMA-specific logic on top of that? It'd be nice to support BLOOM / RWKV / etc models encoded in GGML-format files without having to copy the entire loader.

iacore:
> Do you have an ETA for when you'll be done with the separate crate?
1-2 weeks.

> Also, do you think it would be possible to restructure the loader so that it can parse any valid GGML file and do LLaMA-specific logic on top of that? It'd be nice to support BLOOM / RWKV / etc models encoded in GGML-format files without having to copy the entire loader.

No. Different models have different hyper parameters.

Currently llama.cpp's model format is ad-hoc. I don't know what is "any valid GGML file"

philpax:
> 1-2 weeks.

Hm, do you think it's worth shipping the version without a crate for now and then we get the crate in for the next release? I'd like to release 0.1 this weekend if possible.

> No. Different models have different hyper parameters.

Yeah, I know - but as far as I know, the tensors themselves are stored the same way (including with names), and it's only really the hyperparameters are different. It seems to me that you could parameterise over the hyperparameter struct and support loading any files that have the same {container, model-specific hyperparameters, vocabulary, tensor data} structure.

philpax:
Updated to the latest `main`

iacore:
I think you shouldn't merge this yet.

> Hm, do you think it's worth shipping the version without a crate for now and then we get the crate in for the next release? I'd like to release 0.1 this weekend if possible.
> 

Yes.

> Yeah, I know - but as far as I know, the tensors themselves are stored the same way (including with names), and it's only really the hyperparameters are different. It seems to me that you could parameterise over the hyperparameter struct and support loading any files that have the same {container, model-specific hyperparameters, vocabulary, tensor data} structure.

I'll try when I get to it

danforbes:
> > Yeah, I know - but as far as I know, the tensors themselves are stored the same way (including with names), and it's only really the hyperparameters are different. It seems to me that you could parameterise over the hyperparameter struct and support loading any files that have the same {container, model-specific hyperparameters, vocabulary, tensor data} structure.

> I'll try when I get to it

Given #136, I think it would be good to define a `GgmlLoader` trait and provide implementations for LLaMA and BLOOM (#85) models. Is there anything I can do to help with this?

philpax:
@iacore are you okay with someone else finishing the PR?

philpax:
Awesome, I can take a look at finishing this on the weekend, or you can try going for it, @danforbes. Just let me know.

danforbes:
@philpax - if you like the approach that I have suggested, I would be happy to take a stab at using it to merge the code from @iacore with that from @hhamud.

philpax:
By all means! Feel free to open a PR that combines both of them. Let us know if you need a hand with anything.

Regarding the implementation: principle sounds good to me. What I'd do is keep the code here more or less the same, but have the code to load the model to do one of the following:

1) take a callback to parse the hyperparameters
2) take a trait (your `GgmlLoader`?) to parse the hyperparameters
3) define separate functions to parse each section of the model, then have the loaders for llama/bloom in terms of those functions

I leave it up to you to figure out what would work best. We'd also like to support other models in the future (like RWKV), so nothing too prescriptive.

hhamud:
> @philpax - if you like the approach that I have suggested, I would be happy to take a stab at using it to merge the code from @iacore with that from @hhamud.

Yeah I'm fine with you taking over #85

iacore:
I've moved the crate to `ggml-loader`.

- `ggml-loader` is now not llama-specific
- llama-specific code is moved to `llama-rs::loader2`
- multi-file loading is still not supported (must be handled separately)
- glue-code for `llama-rs` is not yet written.

KerfuffleV2:
Is help needed here or is this already ready to be merged?

philpax:
Up to @iacore if they have time to finish it, but the main things that require fixing are what they've mentioned - multi-file loading of GGML/GGMJ (GGJT is single-file), and to wire up `ggml-loader` to `llama-rs`.

Once this PR is done, we can merge it in, fix the conflicts with #141, and then we're good to go.

KerfuffleV2:
This isn't the approach I would have taken but personally I think it's super important to get support for the latest format. Right now it feels bad that I can't recommend llama-rs to people.

Do you know of any examples of multi-part `GGML` and `GGMF` (I assume you mean rather than `GGMJ`) models that could be used for testing?

philpax:
Er, sorry, yes, I keep making that mistake.

For multi-part: not sure if any are floating around out there. I just took the original LLaMA 13B PTH and converted it using the Python script, which generated two GGMF files.

philpax:
@KerfuffleV2 Were you planning on taking this across the finish line? I'm tempted to do it myself so that we can get the PRs unblocked

danforbes:
I think I can try to pull some of this implementation into my branch. I'm taking a slightly different approach where the `InferenceSession` and `Model` are templatized with respect to a `Context` trait, for which there will be an initial GGML implementation.

iacore:
> Up to @iacore if they have time to finish it, but the main things that require fixing are what they've mentioned - multi-file loading of GGML/GGMJ (GGJT is single-file), and to wire up `ggml-loader` to `llama-rs`.
> 
> Once this PR is done, we can merge it in, fix the conflicts with #141, and then we're good to go.

I probably won't support old multi-file format. Using existing code to load, or provide a link to the C++/Python converter script is good enough.

To wire up `ggml-loader` to `llama-rs`, maybe next week.

KerfuffleV2:
I was personally just trying to find out what's up with it. I didn't want to step on anyone's toes.

The way I'd want to implement this would be to have something that just scans the files and builds up a representation of all the metadata (like types of tensors, offsets/lengths, and so on) in advance of actually loading the data itself.

I actually have been thinking about writing something like that as a separate crate though, just something that can read various model file formats, discover all the information and possibly load/convert/quantize it.

danforbes:
> I didn't want to step on anyone's toes.

You mean you...weren't trying to cause a kerfuffle? Sorry, I could not help myself 😂 10/10 great GitHub handle 💯

philpax:
> I think I can try to pull some of this implementation into my branch. I'm taking a slightly different approach where the `InferenceSession` and `Model` are templatized with respect to a `Context` trait, for which there will be an initial GGML implementation. I will credit @iacore in the respective commit.

Belay that, I'll get this PR good to go

philpax:
> I was personally just trying to find out what's up with it. I didn't want to step on anyone's toes.
> 
> The way I'd want to implement this would be to have something that just scans the files and builds up a representation of all the metadata (like types of tensors, offsets/lengths, and so on) in advance of actually loading the data itself.
> 
> I actually have been thinking about writing something like that as a separate crate though, just something that can read various model file formats, discover all the information and possibly load/convert/quantize it.

That would be quite nice to have (especially as a tool for examining models). We can consider using that if you end up building it, but I'd like to get this in first.

KerfuffleV2:
> I'd like to get this in first.

I 100% agree, getting GGJT loading capability as fast as possible is the most important thing.

Other improvements can come later.

danforbes:
> Belay that, I'll get this PR good to go

LOL good because I was starting to get frustrated 😅

KerfuffleV2:
Uhh, this is tangentially related. Not really sure where to put this question.

I've actually started work on my generic loading stuff. One thing that would make life a _lot_ easier is dealing with entire files as slices instead of trying to integrate parsers with seeking.

So here's the question: Suppose there was a nifty crate that will just give you tensor metadata for ggmf, ggml, ggjt, maybe Torch, SafeTensors, etc but you had to give it the whole file as a buffer (or possibly you could let it open/mmap the stuff itself behind the scenes). (The `safetensors` crate already has that requirement.) Would that be a problem?

philpax:
OK. I attempted to wire up `ggml_loader`, but it looks like there's more work to be done there - I'll leave that for future (will open an issue for @iacore; the specific issues are documented at the top of the module). For now, it's there, but disabled with a const `if`. I've also turned `mmap` back on. @KerfuffleV2, can I borrow your time to review this so that we can merge this?

> So here's the question: Suppose there was a nifty crate that will just give you tensor metadata for ggmf, ggml, ggjt, maybe Torch, SafeTensors, etc but you had to give it the whole file as a buffer (or possibly you could let it open/mmap the stuff itself behind the scenes). (The `safetensors` crate already has that requirement.) Would that be a problem?

The two issues I see are that

1) you may not be able to read the entire file into memory if you're not using mmap. Realistically, I'm not sure how much of an issue this is given you'll need to have the entire model loaded anyway to do inference... but what happens during the loading process if you're not using mmap? Would you end up having both the loaded-file and model in memory at the same time just before completion? If so, won't that double memory use?
2) How would this work with multipart files? The ecosystem seems to be rapidly moving away from them (which is good!) but for the ones out there, how would we tackle them?

KerfuffleV2:
> @KerfuffleV2, can I borrow your time to review this so that we can merge this?

Sure, but I will have to do that tomorrow morning. I have a weird schedule and it's basically my bed time. Was there anything specific you wanted me to look at?

***

> you may not be able to read the entire file into memory if you're not using mmap. 

I'm just talking about mapping it, not necessarily reading it. For the first stage of this, it'll just be scanning metadata so a pretty small amount of the actual file will get read (via `mmap` or otherwise). 

The problem with using a normal file reader is it really complicates parsing, because I want to skip over all the tensors but something like a `nom` parser can't seek. It has a streaming iterator version, but I really don't want to read through the whole file just to discard it.

> but what happens during the loading process if you're not using mmap?

If you don't want to load the actual tensors from the mmap, you could just close it at that point.

> Would you end up having both the loaded-file and model in memory at the same time just before completion?

Well, just `mmap`ping something doesn't use any memory. Once you start looking at the data, it gets pulled in and may count as actual memory. The way I'm envisioning this working is that the first main feature it will have is to just gather all the information about the model: the type of files, start/end offsets of tensors, types, stuff like that. Once you have that, you could just open the file normally, seek to where the tensors are, read in the data, etc.

In the future, I'd like to do other fun stuff like you can take the list of tensors and say "Hey, I want any `Float32` tensors quantized to `GGML_Q4_1` and stuff like that, or have it actually load the tensors, combine ones that are split up, etc.

> How would this work with multipart files?

I don't think that part really has anything to do with `mmap` or not. There are a number of ways to deal with that problem, though. What I've been leaning toward is having the consumer (would be llama-rs or whatever) pass in all the parts in order, if there are multiple parts. Then my crate can deal with figuring out where stuff is.

What I was thinking of is having a pretty generic type that can be passed with each file entry which upstream stuff could put whatever in: could be a file id, could be the whole filename as a string, whatever. Then when you get the metadata you'd get something like a `MultiPartTensor` where you get the start/end offsets and other metadata + that id which can be used to relate it back to the thing it originally came from.

Maybe in the future it can take care of combining the tensors for you but that won't be the first release.

philpax:
It's also *very* past my bed time, so I'll have to respond to you in detail later. For the review, I just want a quick sanity check to make sure there's no obvious issues with the default path. I've tested GGML/GGJT models on Windows with both mmap on and off, so I don't foresee any issues, but you never know.

KerfuffleV2:
@philpax Interestingly, I noticed we're on similar schedules even though it's probably for very different reasons.

I have good news and bad news. GGML and GGMF seem to work perfectly, also exactly the same result (as expected) with a seed using `mmap` vs no `mmap`. I have both versions of the same model and there's also no difference between GGML and GGMF.

However, GGJT doesn't seem to work on the models I have.

`gpt4-x-alpaca-13b-ggml-q4_1-from-gptq-4bit-128g.bin` gives:

    thread 'main' panicked at 'Could not load model: UnsupportedElementType(4)', llama-cli/src/cli_args.rs:322:10

`ggml-vicuna-13b-4bit.bin` "apparently" loads the model but just sits there once it's loaded. I say apparently because:

    2023-04-20T07:37:29Z INFO  llama_cli::cli_args] Model fully loaded! Elapsed: 25ms

25ms? I don't doubt your optimization abilities but that's a little faster than I would expect. :) I'm guessing it's somehow just skipping over the data and not loading those tensors at all.

I haven't looked at the actual code to try to figure out what's wrong. If you need me to help with that part, let me know.

philpax:
I've been testing with `ggml-vicuna-13b-4bit-rev1.bin` and can confirm that the model loading is "instant", but the actual generation takes much longer to start with mmap on. With mmap off, the loading takes as long as usual, but generation starts on time. That's actually why I added `--no-mmap` - I think the behaviour is just not great on Windows, which matches with what people have mentioned in upstream. Will need to test with rev0.

> gpt4-x-alpaca-13b-ggml-q4_1-from-gptq-4bit-128g.bin

Interesting. Here's the relevant code:

https://github.com/iacore/llama-rs-1/blob/69f355b999c11cf670348e65eb1d3a98aa52ff0f/ggml-loader/src/util.rs#L37-L45

I think we need to look at upstream to figure out what their element types are again. What a headache.

I'll have a look at this after work, but feel free to poke around yourself if you want!

KerfuffleV2:
> I've been testing with `ggml-vicuna-13b-4bit-rev1.bin` and can confirm that the model loading is "instant", but the actual generation takes much longer to start with mmap on.

My apologies, you're right. I just didn't wait long enough. That one is actually fine, sorry for the false alarm!

***

If it helps, these are the types (copied from a pretty fresh llama.cpp) from my WIP loader stuff:

```rust
// AKA ftype
pub enum FloatType {
    F32 = 0,
    F16 = 1,
    Q4_0 = 2,
    Q4_1 = 3,
    Q4_1_F16 = 4,
    Q4_2 = 5,
    // Items below this point may get renumbered.
}

pub enum TensorType {
    F32 = 0,
    F16 = 1,
    Q4_0 = 2,
    Q4_1 = 3,
    Q4_2 = 4,
    Q8_0 = 5,
    // Items below this point may get renumbered.
    I8,
    I16,
    I32,
}
```

I guess that would correspond to `ftype` or `f16_`. So it would be `Q4_1_F16` or from the C code:

```c
// tok_embeddings.weight and output.weight are F16
LLAMA_FTYPE_MOSTLY_Q4_1_SOME_F16 = 4, 
```

Fun name, eh? I guess you can expect some tensors to be in F16 format. Maybe it will just get handled automatically with the other stuff if it doesn't bail there.

(I'll probably wait to let you look at it since I want to work on the other thing but if you end up wanting me to specifically try to fix it, just let me know.)

philpax:
What a headache :(

Ok, I'll take a look at this later.

---

Regarding our previous discussion - yeah, not at all concerned about the mmap stuff (your proposed solutions to everything else are fine), more so about the case where mmap isn't available and you need to read the model into memory. The crate we use is tested on x86-64 Linux and Windows, but doesn't seem to test on macOS (either x86-64 or aarch64). I'm sure it'll work fine (I'll test soon on my work laptop), but there might be odd platforms where you don't have access to mmap... someone might try running llama-rs on bare metal (with sufficient std stubbing), for example!

KerfuffleV2:
I lied, I actually looked at the code. One issue I see is it seems like the individual tensor type and the `ftype` or whatever you want to call it is treated the same. I'm not sure if they were the same at one point, but currently they aren't the same enum. 

Model file type: https://github.com/ggerganov/llama.cpp/blob/f7d05095b404b5500b4a702ea16f67fc22446e49/llama.h#L68

Tensor type: https://github.com/ggerganov/llama.cpp/blob/f7d05095b404b5500b4a702ea16f67fc22446e49/ggml.h#L201


***

> but there might be odd platforms where you don't have access to mmap...

Yeah, that's the kind of thing I was talking about. You'd pretty much just have to only support platforms with a working mmap implementation. That should be 100% of unix-type OSes like Mac, Android, etc. Windows also supports it. So even if one specific crate has issues, I'm sure that's not too much of a barrier and of course my side would just see a `&[u8]` so it wouldn't care what mmap crate that came from.

People trying to run on bare metal would be out of luck though, at least initially (unless they set up MMU stuff to simulate mmaping and could pass something a buffer, or were willing to just load the entire file into memory). Would that be acceptable?

iacore:
> edit: I also sent you the links to those models on Discord in case you want to grab them and try to reproduce the problem.

Is there a Discord for llama-rs?

KerfuffleV2:
> I think we would want a way to copy into memory, regardless.)

I think `llama.cpp` deals with that by `mlocking` it (or at least allowing the option to do so). 

But, uhh... Are you actually running inference from `mmap`ped data? Do we even have the capability to do that currently? 

iacore tried with the set_data thing but it caused segfaults if I remember.

If I had to take a guess I'd say the data is still mapped during inference for some reason so the OS is trying to hold onto it, which is causing memory pressure.

If you look at memory stats while it's running, does that process seem like it's using more than you'd expect? The memory usage between mmaped/non should be the same (at least if I'm correct about not actually doing any inference from mmaped data).

philpax:
> But, uhh... Are you actually running inference from `mmap`ped data? Do we even have the capability to do that currently?
> 
> iacore tried with the set_data thing but it caused segfaults if I remember.

Yup! That was fixed in #129, which I subsumed into this PR. The tensors are very much populated from the mmap and not from memory by default for GGJT.

The memory usage is much lower with mmap enabled, and the tensor loading is "instant."

> I don't think phrasing it like that is accurate. `mmap` isn't making anything resident in memory by itself.
> 
> Some stuff _might_ become resident when it's used, but my thing is only going to be reading the metadata so the actual parts of the file that would get read into memory are really small, probably a couple MB at most.

Fair enough. By "make resident" I mean that the file has to be in memory in _some_ form; while that will be a mmap in the majority case (and thus practically free), I'm still a little uncomfortable with degrading to having to load the entire file if mmap is not available for whatever reason.

---

Oh: I just thought of a more realistic scenario - WASM. No mmap there, and it's likely to be an interesting inference platform in future. Right now, it's constrained by the 4GB memory space, but that is not necessarily an issue if you're loading a smaller model.

I can definitely see the advantages of the approach - it's much simpler conceptually and allows for better decoupling of loading - but not all platforms can pay the cost :/

For those constrained platforms, the model could potentially be loaded from a loader optimised for that platform (i.e. custom-packed weights). Hard to say if that's the right approach without knowing what those deployments will look like.

iacore:
WASM very much has mmap.

philpax:
...are you sure about that? I don't think that's the case for the web or for WASI.

iacore:
WASM has Blob and ArrayBuffer, which acts like mmaped file.

KerfuffleV2:
> WASM has Blob and ArrayBuffer, which acts like mmaped file.

You'd still need something to actually mmap the file and pass the binding.

I think you're technically correct though, there isn't really a practical difference. I mean, if someone is trying to run inference on a model in a browser or something that doesn't mmap, you'd have to load the model data anyway. So you'd probably already have a great big buffer with the whole model.

You also probably really wouldn't want to do stuff like try to quantize the model in WASM.

philpax:
> WASM has Blob and ArrayBuffer, which acts like mmaped file. Reading a "file" with POSIX API is more janky.

Yes, but support for multi-memory is limited in all toolchains, so you can't map a file in.

> I mean, if someone is trying to run inference on a model in a browser or something that doesn't mmap, you'd have to load the model data anyway. So you'd probably already have a great big buffer with the whole model.

Not quite - you can stream the file in so that you don't need to have the entire model blob. That would be compatible with the streaming readers we currently have.

> Also, this branch is done. Anyone care to check it out?

Oh, cool! Sorry, didn't see this (was busy the last two days). Looking over it now...

philpax:
@iacore Alright, looks good, but:

1) Do we need to keep loader1 around? Can we replace it entirely with loader2?
2) As mentioned by @KerfuffleV2, `gpt4-x-alpaca-13b-native-4bit-128g.bin` doesn't load: `Could not load model: HyperparametersF16Invalid { ftype: 4 }`. Is that something you would be able to fix?

iacore:
>     1. Do we need to keep loader1 around? Can we replace it entirely with loader2?

keep the multi-file part

>     2. As mentioned by @KerfuffleV2, `gpt4-x-alpaca-13b-native-4bit-128g.bin` doesn't load: `Could not load model: HyperparametersF16Invalid { ftype: 4 }`. Is that something you would be able to fix?

That seems like a separate issue

KerfuffleV2:
> That seems like a separate issue.

It's a LLaMA model, not a RWKV model. RWKV models wouldn't be expected to work currently.

The problem is the code is treating the model `f16` type the same as the tensor type, and they aren't the same. See the top part of: https://github.com/rustformers/llama-rs/pull/125#issuecomment-1515966952

llama-rs/src/vocabulary.rs
```diff
@@ -65,9 +59,7 @@ impl Vocabulary {
         self.max_token_length = self.max_token_length.max(content.len());
         self.id_to_token.push(content.clone());
         self.id_to_token_score.push(score);
-        if self.token_to_id.insert(content, id).is_some() {
```
iacore: this is sanity check, maybe don't delete this

llama-rs/src/loader2.rs
```diff
@@ -62,7 +62,7 @@ pub(crate) fn load(
     ggml_loader::load_model_from_reader(&mut reader, &mut loader)
         .map_err(|err| LoadError::from_ggml_loader_error(err, path.clone()))?;
 
-    Ok(loader.model.expect("model should be initialized"))
+    loader.model.ok_or(LoadError::ModelNotCreated { path })
```
iacore, line 65: the way I wrote the code, `Err` is actually impossible here.

llama-rs/src/loader2.rs
```diff
@@ -110,9 +110,9 @@ impl<F: FnMut(LoadProgress)> ggml_loader::LoadHandler<LoadError, BufReader<&File
             Ok(id) => id,
             Err(err) => return ControlFlow::Break(LoadError::InvalidIntegerConversion(err)),
         };
-        self.vocab
-            .push_token(id, token, score)
-            .expect("vocab should be valid");
+        if let Err(err) = self.vocab.push_token(id, token, score) {
```
iacore, line 113: again, given the API usage, this is impossible

philpax:
> keep the multi-file part

I'm leaning towards disabling multi-file loading entirely and using loader1 to create a converter from GGML/GGMF multipart to GGJT. Keeping two loaders around just to handle a fairly niche use-case doesn't seem worth it to me.

> That seems like a separate issue.

Yeah, as Kerfuffle said, that's an issue with an existing LLaMA model (you can google that filename). That being said, I'm thinking we should address that in another PR.

llama-rs/src/vocabulary.rs
philpax: I added the sanity check - your code wasn't doing it before: https://github.com/rustformers/llama-rs/pull/125/commits/9b908ae57a77f0a134fd2720d66f024f06286ce4#diff-437de44ac54122eab1784619577c65d5d869088c21979e99292d5a089768bceeR68

I removed the check because my GGML Alpaca 7b q4 had several duplicate tokens for ' ' with both loader1 and 2. If models are out there with duplicate terms in the vocabulary, we'll need to support them.

llama-rs/src/loader2.rs
philpax: It's possible if there are no tensors at all in the model (e.g. `tensor_buffer` never gets called). Users should be aware of that.

llama-rs/src/loader2.rs
philpax: Hm. Sure. It mattered more when there was an error for duplicate tokens, but in this case it would be a breaking of the loading invariants if we encounter an ID out of step. I'll change that back to a panic.

llama-rs/src/vocabulary.rs
iacore: interesting

iacore:
should we merge this now?

philpax:
Give me a sec, just going to make multifile loading on loader2 panic for now and then I'll merge

iacore:
Please make it instruct using `env GGML_LOADER=1` for multi file loading

KerfuffleV2:
Why use weird environment variables instead of just having a commandline option?

Since multi-part models are a rare case, it seems like it makes the most sense to have single file as the default and specifying (or setting to multi-file mode) with a CLI option.

philpax:
loader2 is the default. I was considering making it a CLI option myself, but with

> I'm leaning towards disabling multi-file loading entirely and using loader1 to create a converter from GGML/GGMF multipart to GGJT. Keeping two loaders around just to handle a fairly niche use-case doesn't seem worth it to me.

my plan is to remove support entirely for multipart + loader1 and relegate it to that separate tool. I'm guessing you haven't found that many multipart models in the wild, right?

philpax:
> The current loader2 code doesn't detect multiple files; simply assumes it's only 1 file.
> 
> Please make it instruct using `env GGML_LOADER=1` for multi file loading.

Are you OK with me removing `LoadHandler::load_multipart` entirely?

KerfuffleV2:
> I'm guessing you haven't found that many multipart models in the wild, right?

I didn't try.

iacore:
> > The current loader2 code doesn't detect multiple files; simply assumes it's only 1 file.
> > Please make it instruct using `env GGML_LOADER=1` for multi file loading.
> 
> Are you OK with me removing `LoadHandler::load_multipart` entirely?

Yes.

philpax:
Alright fellas, last chance to object before merge. After this, I'm going to

- merge #84 
- use it to build a utility to convert multipart models to singlepart GGJT; I'll likely rename `ggml_loader` to `ggml_format` and write code for writing GGJT from hyperparameters + vocab + tensors
- create an issue for loading GPT4-X
- bring #141 up to date

llama-rs/src/vocabulary.rs
```diff
@@ -26,41 +24,23 @@ pub struct Vocabulary {
...
     /// Add a token to the vocabulary.
     ///
     /// The token added must have `id` directly after the last token in the vocabulary.
-    pub fn push_token(
-        &mut self,
-        id: TokenId,
-        content: Token,
-        score: TokenScore,
-    ) -> Result<(), AddTokenError> {
+    pub fn push_token(&mut self, id: TokenId, content: Token, score: TokenScore) {
```
iacore, line 31: should probably call this `push_token_unchecked` or something

philpax: I think panicking here is acceptable - if it panics, there's a bug in the loader and it should be fixed as soon as possible. It's not something that could be meaningfully recovered from as an end-user.

iacore: Can you change this to `pub(crate)` and change the function name?

philpax: Sure. I'll keep the function name the same (there's not really a convention around marking functions that can panic in Rust, as far as I know), but I'll mark it `pub(crate)` and add a note to the doc comment about its panic-safety.

iacore: it's usually called `*_uncheck` or something

PR #128, opened 2023-04-11
Change Rust-side GGML view ops to take element rather than byte arguments.

KerfuffleV2:
I think this is kind of a no-brainer change because it I can't think of a reason one would want to view a tensor except at the granularity of the elements. Source: Trust me bro.

I'd like to propose this as something that comes in before the 0.1 release because it would be a breaking API change.

### DIFF
--- a/ggml/src/lib.rs
+++ b/ggml/src/lib.rs
@@ -235,6 +235,7 @@ impl Context {
 
     /// Creates a 1D view over `a`.
     pub fn op_view_1d(&self, a: &Tensor, ne0: usize, offset: usize) -> Tensor {
+        let offset = offset * a.element_size();
         let tensor = unsafe {
             ggml_sys::ggml_view_1d(self.ptr.as_ptr(), a.ptr.as_ptr(), usize_to_i64(ne0), offset)
         };
@@ -250,6 +251,10 @@ impl Context {
         nb1: usize,
         offset: usize,
     ) -> Tensor {
+        let elsize = a.element_size();
+        let offset = offset * elsize;
+        let nb1 = nb1 * elsize;
+
         let tensor = unsafe {
             ggml_sys::ggml_view_2d(
                 self.ptr.as_ptr(),
@@ -275,6 +280,11 @@ impl Context {
         nb2: usize,
         offset: usize,
     ) -> Tensor {
+        let elsize = a.element_size();
+        let offset = offset * elsize;
+        let nb1 = nb1 * elsize;
+        let nb2 = nb2 * elsize;
+
         let tensor = unsafe {
             ggml_sys::ggml_view_3d(
                 self.ptr.as_ptr(),
--- a/llama-rs/src/lib.rs
+++ b/llama-rs/src/lib.rs
@@ -1130,9 +1130,6 @@ impl Model {
         let n_past = session.n_past;
         let n_threads = params.n_threads;
 
-        let memk_elsize = session.memory_k.element_size();
-        let memv_elsize = session.memory_v.element_size();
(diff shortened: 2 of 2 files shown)


philpax:
Hm, this is definitely better but I fear that it'd make porting code much more confusing (especially because the functions would look the same but behave differently). Maybe it's worth prototyping a `Size::{Elements, Bytes}` enum that makes it explicit? e.g. 

```rust
pub fn op_view_1d(&self, a: &Tensor, ne0: usize, offset: Size) {
  let offset = size.byte_count(self.element_size());
}
```

```rust
&ctx0.op_view_1d(
    &session.memory_k,
    (n_past + n) * n_embd,
    ggml::Size::Elements(il * n_ctx * n_embd),
),
```

KerfuffleV2:
I guess it's worth asking: What is the `ggml` crate supposed to be? There's already `ggml-sys` which is a direct low level binding. Is `ggml` supposed to be a very direct C-in-Rust that's a little easier to use than `ggml-sys` or is it aiming to be a crate for "Using GGML from Rust in the Rust way"?

If it's the first one, then making users multiply everything by the element size even though you'd _never_ want to do anything else is understandable, I guess. As for the enum idea, I think it's a little better than the status quo but it still seems weird to me to have to type in a bunch of extra stuff to do what basically everyone is already going to want to do 99% of the time.

My own personal philosophy is not to try to work too hard to prevent people from shooting themselves in the foot, when it comes at a cost of convenience/performance. (I also dislike it when I want to do something and I can't or it's harder than it needs to be because someone else is trying so hard to protect me from myself.)

Hopefully this doesn't sound too harsh, I'm a pretty opinionated person and ironically I hate it when I run into things that are opinionated too...

philpax:
> I guess it's worth asking: What is the ggml crate supposed to be? There's already ggml-sys which is a direct low level binding. Is ggml supposed to be a very direct C-in-Rust that's a little easier to use than ggml-sys or is it aiming to be a crate for "Using GGML from Rust in the Rust way"?

Great question. It's something we've been trying to settle on as well, but I'd lean more towards "mildly nicer Rust binding" than "a fully Rust solution", at least for our immediate goals. I think the latter would be nice to have and perhaps we should even aim for it, but it's not necessarily what we need for this stage (especially with the explosion of projects based on `llama.cpp` that we need to keep an eye on)

> As for the enum idea, I think it's a little better than the status quo but it still seems weird to me to have to type in a bunch of extra stuff to do what basically everyone is already going to want to do 99% of the time.

Totally agree - it's more that we'll be porting a lot of code verbatim from C++, and I don't want something that looks the same as the C code to behave differently. That's going to be a confusing gotcha, even if it's made with good intentions in mind.

> My own personal philosophy is not to try to work too hard to prevent people from shooting themselves in the foot, when it comes at a cost of convenience/performance. (I also dislike it when I want to do something and I can't or it's harder than it needs to be because someone else is trying so hard to protect me from myself.)
> 
> Hopefully this doesn't sound too harsh, I'm a pretty opinionated person and ironically I hate it when I run into things that are opinionated too...

No worries at all, I get it. I'm not saying you're wrong - I think this change would be sensible for a more general-purpose solution - but with the current state of affairs, I'd like to prioritise clarity for porting C++ code.

---

With that being said, I'd like to treat GGML as an implementation detail in the long run and describe models in a more Rust-y way where it's obvious that it doesn't map 1:1 to the original GGML code. I'd like to hear your thoughts on #137, which is where I think we should be heading. For that, I'd work over elements and *never* expose raw bytes, except if it's relevant to getting data in or out.

PR #129, opened 2023-04-12
feat: mmapped ggjt loader

jon-chuang:
Fixes the issues in https://github.com/rustformers/llama-rs/pull/125

Improvements:
- Loading **7B Vicuna (q4)**: 
  - **default**: warm start: **1785ms**, cold start: **2618ms**
  - **`--features="mmap"`**: warm start: **7ms**, cold start: **38ms** 
- Loading **13B Vicuna (q4)**:
  - **default**: warm start: **4833ms**, cold start: **5905ms**
  - **`--features="mmap"`**: warm start: **9ms**, cold start: **33ms** 

So we get a 250X-500X speedup! Higher than the advertised 10-100x :)

@iacore

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,46 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.2.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "342258dd14006105c2b75ab1bd7543a03bdf0cfc94383303ac212a04939dff6f"
+dependencies = [
+ "anstyle",
+ "anstyle-parse",
+ "anstyle-wincon",
+ "concolor-override",
+ "concolor-query",
+ "is-terminal",
+ "utf8parse",
+]
+
+[[package]]
+name = "anstyle"
+version = "0.3.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


jon-chuang:
Btw, the best part about this is that the OS page cache is reused between llama.cpp and llama-rs!

jon-chuang:
Another optimization - we don't have to allocate memory for `new_tensor_{}d` for mmap case.

iacore:
The `llama-loader` crate is still WIP (can only load GGJT, and isn't used by `llama-rs`). I think it's better for you to rebase this on https://github.com/rustformers/llama-rs/pull/129/commits/cc846aee76b09aa434430ba0ccf9d35c5fe9cbdc

jon-chuang:
> The cost of page fault is not paid until first access, so 38ms is definitely not right

I suppose. But it requires measuring the inference time.

At the least, the user can interact right away, which is a plus.

jon-chuang:
> can only load GGJT, and isn't used by llama-rs

We should add tests to show proper loading for all these formats. 

We need some simple models, we can generate them with a simple script. I'd also like to implement a way to dump the loaded model into a file.

philpax:
I would suggest making this a PR to iacore's work, so that we can merge #125 and have this included.

iacore:
I think this should be merged first. #125 has less user-facing features than this.

philpax:
I've subsumed this into #125 - thanks for the PR, awesome to see `mmap` working 🚀

PR #135, opened 2023-04-13
Split up modules

philpax:
This is a little overdue I think, and it'll cause problems for the other PRs, but it also makes it much easier to maintain.

I've made a few controversial changes in the last three commits. The rest are pretty straightforward.


llama-cli/src/main.rs
```diff
@@ -73,8 +72,8 @@ fn infer(args: &cli_args::Infer) {
 
 fn dump_tokens(args: &cli_args::DumpTokens) {
     let prompt = load_prompt_file_with_prompt(&args.prompt_file, args.prompt.as_deref());
-    let (_, vocabulary) = args.model_load.load();
-    let toks = match vocabulary.tokenize(&prompt, false) {
+    let model = args.model_load.load();
```
setzer22, line 75: I agree with the change. I had considered doing this a few times before. Since both the model and the vocabulary are meant to be immutable after creation, bundling them into the same struct will hardly cause issues.

llama-rs/src/inference_session.rs
```diff
@@ -0,0 +1,574 @@
...
+use crate::{
+    util::mulf, EvaluateOutputRequest, InferenceError, InferenceParameters, Model, TokenId,
+    TokenUtf8Buffer, EOT_TOKEN_ID,
+};
+
+// The size of a scratch buffer used for inference. This is used for temporary
+// storage of intermediate results during inference.
+//
+// The specific value was copied from `llama.cpp`.
+const SCRATCH_SIZE: usize = 512 * 1024 * 1024;
```
setzer22, line 16: I think llama.cpp figured out a proper way to compute this value. We should have a look at this. Not in this PR of course :+1:

setzer22 (approved):
Everything looks good. I think the way modules are split makes sense :+1: 

Hopefully this wouldn't cause too many conflicts with existing PRs? e.g. #125 comes to mind

llama-rs/src/inference_session.rs
philpax: image

https://github.com/ggerganov/llama.cpp/blob/82d146df9b43cf677e0dbce20b03cf864958a0cc/llama.cpp#L45-L57

philpax:
> Hopefully this wouldn't cause too many conflicts with existing PRs? e.g. https://github.com/rustformers/llama-rs/pull/125 comes to mind

Yeah, I'm a little concerned about that one myself, but I don't think it should be too bad - it'll mostly just be discarding the changes to `loader.rs` here. The others shouldn't be too hard to accommodate.

PR #138, opened 2023-04-14
Sync to GGML version as of 20230414 06:00 AM UTC.

KerfuffleV2:
This also exposes the unary and binary map operations in the `ggml` crate.

Doesn't appear to be any inference related changes in `llama.cpp` itself so nothing to worry about there.

### DIFF
--- a/ggml-sys/ggml/CREDITS.txt
+++ b/ggml-sys/ggml/CREDITS.txt
@@ -1,4 +1,6 @@
-Vendored version: https://github.com/ggerganov/llama.cpp/commit/180b693a47b6b825288ef9f2c39d24b6eea4eea6
+Vendored version: https://github.com/ggerganov/llama.cpp/commit/f4d277ae17247ee51129ef1a9ff74d377cc90b1b
+For convenience, changes involved in this sync:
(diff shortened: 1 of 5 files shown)


KerfuffleV2:
I was thinking there might be an argument for using the typedefs from the C side in the bindings:

```rust
pub type ggml_unary_op_f32_t = ::std::option::Option<
    unsafe extern "C" fn(arg1: ::std::os::raw::c_int, arg2: *mut f32, arg3: *const f32),
>;
pub type ggml_binary_op_f32_t = ::std::option::Option<
    unsafe extern "C" fn(
        arg1: ::std::os::raw::c_int,
        arg2: *mut f32,
        arg3: *const f32,
        arg4: *const f32,
    ),
>;
```

However, that would require passing the function pointers in as an `Option` which is kind of pointless, since if you deliberately pass `NULL` as a function pointer to the map operations... Well, you might have been dropped on your head back in the formative years.

Also, they're rather less descriptive than the manual ones. I wouldn't expect those types to change in the future.

philpax (approved):

philpax:
Yeah I wouldn't worry too much about the typedef stuff, can figure that out later if needed

PR #148, opened 2023-04-21
Add color_eyre for cli error handling

tehmatt:
We have nice error messages via `thiserror` - let's use them! 
This PR uses `color_eyre` to get fancier error messages, including tracking context around errors to make debugging easier.

There's probably more error handling we could convert to this approach, but this is decent starting point.


tehmatt:
It seems like there's a decent amount of work going on in #125, but this looks like it won't cause any merge conflicts.

philpax:
Awesome, thank you for your contribution! Looks great 🚀

PR #156, opened 2023-04-25
fix #155 - multi-line text entry

### LINKED ISSUE #155: Feature request: multi-line text entry
Llama.cpp's chat mode allows entering multiple lines of text by adding a trailing `\` to each one.  I think this would be a nice feature to have, either by copying the backslash functionality over or by using shift+enter like many chat clients.
(I have this partly working already but I want to make sure there's support before going any farther.)

suzuke:
I have implemented multi-line input. 
May you please help me check if it is feasible?

### DIFF
--- a/binaries/llm-cli/src/cli_args.rs
+++ b/binaries/llm-cli/src/cli_args.rs
@@ -281,7 +281,6 @@ impl ModelLoad {
     pub fn load<M: llm::KnownModel + 'static>(&self) -> Result<Box<dyn Model>> {
         let now = std::time::Instant::now();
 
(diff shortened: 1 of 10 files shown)


iacore:
readline history will be messed up. Is it possible for you to use multi-line readline?

suzuke:
I have tried to make corrections, but the implementation is not significantly different from  #157

philpax:
Thanks for the PR! Sorry that you felt you had to close it after the refactor - I'll try to do a better job of communicating / managing changes like that in the future.

Issue #53, opened 2023-03-21
Better multiline support in interactive mode


0xAlcibiades:
~~For longer form interaction, it'd be really nice to add 1) lookback/context 2) better formatting/line returns.~~

@philpax edit: Lookback exists in chat mode. The main thing that this issue now covers is multiline support in interactive mode: you should be able to supply multi-line prompts and have their output rendered correctly.

philpax:
There's overlap with #76 here. We'll need to create a new mode for this and clearly delineate what they do.

philpax:
Lookback/context exists in the new experimental chat model. I'll update this issue to refer to multi-line prompts in interactive mode.

philpax:
Done in #157 🚀

PR #157, opened 2023-04-26
Add multiline support to chat

dustletter:
This addresses #53 / the ill conceived #155.
Implementing this using rustyline allows you to backspace over newlines, although I'm not sure if that's worth it.  Otherwise I like the approach in #156 too!

I *think* the code for shift+enter is correct, but either my terminal isn't sending the right keys or rustyline isn't detecting them.


philpax:
Hm, I'll have to check out both this and #156! Will try out when I can :)

philpax:
Thanks for the PR!

Shift-enter doesn't work for me either (Windows), but that's fine; it can be figured out later.

pdufour:
Hi @philpax just wanted to see if shift enter should be working on the latest main or that is still an issue? Thanks.

PR #159, opened 2023-04-26
fix: Proper error message when the model file doesn't exist

tehmatt:
Present the error from reading the file, rather than an error about multipart support.


philpax:
Thanks for both PRs, much appreciated!

PR #160, opened 2023-04-27
mmap with MAP_POPULATE

iacore:
Linux man page says about MAP_POPULATE

       MAP_POPULATE (since Linux 2.5.46)
              Populate (prefault) page tables for a mapping.  For a file
              mapping, this causes read-ahead on the  file.   This  will
              help  to reduce blocking on page faults later.  The mmap()
              call doesn't fail if the mapping cannot be populated  (for
              example,  due  to limitations on the number of mapped huge
              pages when using MAP_HUGETLB).  Support  for  MAP_POPULATE
              in  conjunction  with  private mappings was added in Linux
              2.6.23.

### DIFF
--- a/llama-rs/src/loader.rs
+++ b/llama-rs/src/loader.rs
@@ -16,7 +16,6 @@ use ggml_format::{
     util::{has_data_left, read_bytes_with_len, read_f32, read_i32, read_u32},
     ContainerType,
 };
-use memmap2::Mmap;
 
 pub(crate) fn load(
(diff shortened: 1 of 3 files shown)


philpax:
Thanks for the contribution! Out of curiosity, was there any particular reason you moved the `mmap` call to a `util` function? Were you planning to use it for something else?

iacore:
It was used twice. I don't want code duplication, where a future change may only change 1 of 2 used locations. That's bad.

Issue #22, opened 2023-03-16
broken build on fedora


sylvain-reynaud:
The build is broken on fedora 37.

full logs : https://gist.github.com/sylvain-reynaud/fe73ccc7edad1f4f98688cb48b1f101c

```
--- stderr
  ggml/ggml.h:177:10: fatal error: 'stddef.h' file not found
  thread 'main' panicked at 'Unable to generate bindings: ClangDiagnostic("ggml/ggml.h:177:10: fatal error: 'stddef.h' file not found\n")', ggml-raw/build.rs:31:10
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

## Fix

On Fedora, to avoid build errors due to the missing `stddef.h` file, you may need to install the following packages:

``` shell
sudo dnf groupinstall "Development Tools" "Development Libraries"
```

Then build with `CPATH` set to the location of the `gcc` headers :
      
``` shell
CPATH="/usr/lib/gcc/x86_64-redhat-linux/12/include/" cargo build --release
```

Tested on my laptop fedora 37 and fedora:latest docker image.

setzer22:
That's odd... Of all the possible distros, I happen to be on Fedora 37! 🤣 

The thing is, I had to do nothing of the sort. I probably already had the C compiler + headers from other projects, but I definitely didn't have to set up `CPATH`, the C code builds for me just fine.

sylvain-reynaud:
@setzer22 ahah, sure, I had tested it:

`docker run -it fedora bash`

Then

```bash
ls /usr/lib/        #no gcc
dnf groupinstall "Development Tools" "Development Libraries"
ls /usr/lib/gcc/x86_64-redhat-linux/12/include/        #installed
git clone https://github.com/sylvain-reynaud/llama-rs
cd llama-rs/
curl https://sh.rustup.rs -sSf | sh
source "$HOME/.cargo/env"
cargo build --release        #still fails
CPATH="/usr/lib/gcc/x86_64-redhat-linux/12/include/" cargo build --release        #success
```

thank you for your fast response btw

sylvain-reynaud:
I just tried by opening a new bash in the previous container, cargo clean and cargo build --release. It is still broken.

setzer22:
Thanks again! Will try to investigate a bit more why exactly those CFLAGS aren't needed on my system 🤔

sylvain-reynaud:
@setzer22 I'm sorry I have no idea. I may take a look this evening.

[...]

PR #162, opened 2023-04-30
Structural Overhaul

danforbes:
Replaces #141. Includes code from #85 and #125.

- GGML is now a submodule and a single crate
- Support for GPT-2 architecture
- Structural refactor into `llm`, `llm-base`, model, and `llm-cli` crates


philpax:
Aha, I was just about to reply to the BLOOM PR!

I would suggest keeping BLOOM in - there's no reason to remove that code if someone can fix it with some work (we'll make an issue for it) - but we'll remove it from the list of features, and make a note on the crate itself.

As for the rest, let me have a look...

README.md
```diff
@@ -1,39 +1,31 @@
 # LLaMA-rs
 
-<!-- markdownlint-disable-file MD026 -->
+This project is a Rust port of
+[llama.cpp](https://github.com/ggerganov/llama.cpp) 🦙🦀🚀
```
philpax, line 4: I'm thinking we'll remove this wording as we grow to accommodate more LLMs. I'll revise the wording on this after this PR lands, so nothing for you to do here - just mentioning it.

danforbes:
I don't see the value of keeping non-working code in the main branch. I would suggest keeping BLOOM in a feature branch until it's working.

README.md
```diff
@@ -43,41 +35,45 @@ Debug builds are going to be very slow.
...
+Clone the repository and then build it with
 
 ```shell
-cargo build --release --bin llama-cli
+git clone --recurse-submodules git@github.com:rustformers/llama-rs.git
+cargo build --release
 ```
 
-The resulting binary will be at `target/release/llama-cli[.exe]`.
+The resulting binary will be at `target/release/llm-cli[.exe]`.
```
philpax, line 57: I think the CLI is called `llm`? The crate is `llm-cli`, but the binary itself is `llm`.

ggml-rs/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "ggml-rs"
```
philpax, line 2: I would keep this as `ggml` - I personally dislike the `-rs` suffix because it remaps to `_rs` in the crate.

ggml-rs/build.rs
```diff
@@ -88,6 +89,16 @@ fn main() {
     }
     build.warnings(false);
     build.compile("ggml");
+
+    let header_path = "./ggml/include/ggml/ggml.h";
```
philpax, line 93: Ah yeah - we originally did this, but moved to pre-generated bindings because not everyone has Clang/LLVM available. (#22)

Unfortunately, I think we'll have to stick with the pre-generated bindings 😦

README.md
danforbes: Good catch!

ggml-rs/build.rs
danforbes: That...is unfortunate 😅

hhamud:
btw could you add me as a git commit as you did when you upstreamed the ggml changes.

philpax:
> I don't see the value of keeping non-working code in the main branch. I would suggest keeping BLOOM in a feature branch until it's working.

Code that isn't kept in the main branch tends to rot, especially if we start making larger changes. As far as I know, BLOOM is done, we just need to figure out what's causing it to fail generation.

It's much easier for someone to pick it up and fix it if it's present; I think it's also a good way to ensure our abstractions work as expected as we increase the number of models.

---

Apart from that, PR looks great! Well done :) Looking forward to merging this, I think it's a huge improvement.

hhamud:
> > I don't see the value of keeping non-working code in the main branch. I would suggest keeping BLOOM in a feature branch until it's working.
> 
> Code that isn't kept in the main branch tends to rot, especially if we start making larger changes. As far as I know, BLOOM is done, we just need to figure out what's causing it to fail generation.
> 
> It's much easier for someone to pick it up and fix it if it's present; I think it's also a good way to ensure our abstractions work as expected as we increase the number of models.
> 
> Apart from that, PR looks great! Well done :) Looking forward to merging this, I think it's a huge improvement.

I personally think we should just merge and place a comment on the issue of it not working. I'd pick it up again but I don't have time for it till next week.

ggml-rs/Cargo.toml
danforbes: All set ✅

danforbes:
I've added back the BLOOM crate and tagged @hhamud as a co-author ✅

philpax:
All that's left are the formatting and restoring bindings generation. Are you OK for me to do that, or do you have changes in the works?

danforbes:
Please go for it - thank you so much 👯

Issue #147, opened 2023-04-20
Support for SalesForce CodeGen


philpax:
https://github.com/salesforce/CodeGen
https://huggingface.co/docs/transformers/v4.28.1/en/model_doc/codegen#transformers.CodeGenForCausalLM

Apparently one of the best for code-generation.

hhamud:
Can you assign this to me

philpax:
GPT-J is very similar to CodeGen, but not quite: https://gist.github.com/moyix/7896575befbe1b99162ccfec8d135566 (h/t @hhamud)

Once GPT-J's merged, we can look at either adding an option to GPT-J or setting up a separate CodeGen model that shares 99% of the logic.

hhamud:
> GPT-J is very similar to CodeGen, but not quite: https://gist.github.com/moyix/7896575befbe1b99162ccfec8d135566 (h/t @hhamud)
> 
> Once GPT-J's merged, we can look at either adding an option to GPT-J or setting up a separate CodeGen model that shares 99% of the logic.

I've noticed that others just use that script to convert codegen to be able to run on code that is based on GPT-J's model.

PR #165, opened 2023-05-01
Implement Codegen Model

hhamud:
Solves #147


hhamud:
Rip another in the dust. GPT-J Implementation works with the codegen models. Complete here #168

PR #168, opened 2023-05-02
GPT-J model implementation

danforbes:
Tested with https://huggingface.co/Kastor/GPT-J-6B-Pygway-ggml-q4_1/blob/main/GPT-J-6B-Pygway-v2-ggml-q4_1.bin


philpax:
About to sleep so I can't look over this properly, but looks good at a glance! 

- Can you rename `Gptj` to `GptJ`?
- Does `llm gptj` work?
- Is there a way to share code amongst the examples? They look pretty similar.

Otherwise, should be fine.

hhamud:
This also solves #165 and its been tested with these models here https://huggingface.co/Guglielmo. Will be closing my PR #165

PR #177, opened 2023-05-05
Speed up model loading

tehmatt:
We were previously cloning too much, causing even `mmap` file loads to be fairly slow.  After this patch, load times for alpaca-65b drop from 300ms minimum to ~6ms on my machine.  Similarly, I now get ~16s load times for `--no-mmap`.

I also snuck in a change to render the spinner less frequently, since `spinoff` apparently throws away spinner state on `update_text()` calls. I'm too lazy to make a second PR and that's arguably related to loading perf, so here it is.


philpax:
Oops, good catch!

PR #178, opened 2023-05-05
Consider current directory if parent directory is blank

tanmaysachan:
Small QOL fix.
Currently when the model is in the current directory, using the model name directly throws awkward "non-specific I/O error".
Better to simply check if we're in the current directory.


philpax:
Well spotted! Totally forgot that some people might have models in the same folder as the executable.

tanmaysachan:
@philpax fixed fmt issues, should be okay for merge.

PR #180, opened 2023-05-05
Add support for newer GGML file types

danforbes:
Tested with https://huggingface.co/xzuyn/Pygmalion-V3-6B-ggml-q5_1/blob/main/ggml-model-q5_1.bin


philpax:
Good to go outside of the doc fail. Maybe use `#[deprecated]` for that?

PR #186, opened 2023-05-06
Codegen Implementation

hhamud:
Completes #165 

## TODO:

- [x] find out how codegen model differs to GPT-J
- [ ] Implement difference
- [ ] Convert codegen HF model to GGML
- [ ] refactor code to share with GPT-J

### DIFF
--- a/binaries/llm-cli/src/cli_args.rs
+++ b/binaries/llm-cli/src/cli_args.rs
@@ -32,6 +32,12 @@ pub enum Args {
         #[command(subcommand)]
         args: BaseArgs,
     },
+    /// Use a GPT-J model
+    #[clap(id = "codegen")]
+    CodeGen {
+        #[command(subcommand)]
+        args: BaseArgs,
+    },
     /// Use a GPT-NeoX model
     #[clap(id = "neox")]
     NeoX {
--- a/binaries/llm-cli/src/cli_args.rs
+++ b/binaries/llm-cli/src/cli_args.rs
@@ -32,7 +32,7 @@ pub enum Args {
         #[command(subcommand)]
         args: BaseArgs,
     },
-    /// Use a GPT-J model
+    /// Use a CodeGen model
     #[clap(id = "codegen")]
     CodeGen {
         #[command(subcommand)]
--- a/crates/models/codegen/src/lib.rs
+++ b/crates/models/codegen/src/lib.rs
@@ -1,15 +1,14 @@
-
-// Ref: https://github.com/ggerganov/ggml/blob/abea4b7/examples/gpt-j/main.cpp
-
 use std::{error::Error, path::Path};
 
 use ggml::Tensor;
 use llm_base::{
-    util, BasicWriteError, EvaluateOutputRequest, FileType, InferenceParameters, InferenceSession,
-    InferenceSessionParameters, KnownModel, LoadError, LoadProgress, Mmap, TensorLoader, TokenId,
-    Vocabulary,
+    ggml, model::common, util, BasicWriteError, EvaluateOutputRequest, FileType,
+    InferenceParameters, InferenceSession, InferenceSessionParameters,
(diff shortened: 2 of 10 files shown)


crates/models/codegen/src/lib.rs
```diff
@@ -0,0 +1,427 @@
...
+    _context: ggml::Context,
+}
+
+unsafe impl Send for CodeGen {}
+unsafe impl Sync for CodeGen {}
+
+impl CodeGen {
+    /// Load the model from `path` with `n_context_tokens` context tokens.
+    ///
+    /// The status of the loading process will be reported through `load_progress_callback`.
```
danforbes, line 47: This is a copy-and-paste error that has been fixed in a branch - please refer to that branch for the documentation format.  https://github.com/danforbes/llama-rs/blob/dfo/chore/models/crates/models/bloom/src/lib.rs#L44-L46

crates/models/codegen/src/lib.rs
```diff
@@ -0,0 +1,427 @@
...
+                //),
+                //n_past,
+                //n_rot,
+                //0,
+            //);
+//
+            //// self-attention store key and value to memory
+            //let vcur =
+                //ctx0.op_transpose(&ctx0.op_mul_mat(&self.layers[il].c_attn_v_proj_w, &current));
+//
```
danforbes, line 213: I don't really love checking this kind of stuff into a `main` branch. Can this be removed? If something needs to be persisted for posterity, maybe it's more appropriate to use an Issue?

crates/models/codegen/src/lib.rs
```diff
@@ -0,0 +1,427 @@
...
+        common::extract_embeddings(output_request, &embd, n_embd, n);
+        common::update_session(session, &ctx0, input_tokens.len(), n);
+    }
+
+    fn vocabulary(&self) -> &Vocabulary {
+        &self.vocabulary
+    }
+
+    fn n_context_tokens(&self) -> usize {
+        self.hyperparameters.n_ctx
```
danforbes, line 325: Is this right? I'm worried this is causing problems with GPT-2, but I'm not sure...
https://github.com/rustformers/llm/issues/167#issuecomment-1537160402

binaries/llm-cli/src/cli_args.rs
```diff
@@ -32,6 +32,12 @@ pub enum Args {
         #[command(subcommand)]
         args: BaseArgs,
     },
+    /// Use a GPT-J model
```
danforbes, line 35:
```suggestion
    /// Use a CodeGen model
```

danforbes (approved):
Awesome! Please make sure the docs are updated, though 🙏🏻

philpax:
How different is this to the original GPT-J implementation? Can the codegen model be implemented by calling into GPT-J with a parameter to use a slightly different computation graph?

I'd really like to avoid any unnecessary duplication if possible, especially as we evolve the models.

crates/models/codegen/src/lib.rs
hhamud: Yeah I will remove that. I'm just in the middle of finding out how to best implement the qkv vector implementations.

crates/models/codegen/src/lib.rs
hhamud: probably not I'll fix that when refactoring. Goal is just to get it working with a converted codegen model

hhamud:
> How different is this to the original GPT-J implementation? Can the codegen model be implemented by calling into GPT-J with a parameter to use a slightly different computation graph?
> 
> I'd really like to avoid any unnecessary duplication if possible, especially as we evolve the models.

That's more or less what I'm thinking of doing but the stage I'm currently at is to try and implement codegen's treatment of the QKV vectors.

ralyodio (approved):

hhamud:
No longer useful

PR #191, opened 2023-05-07
Replace custom ModelArchitecture impl with idiomatic traits

danforbes:


crates/llm/src/lib.rs
```diff
@@ -103,59 +107,67 @@ pub enum ModelArchitecture {
     /// [GPT-NeoX](llm_neox)
     NeoX,
 }
-impl Display for ModelArchitecture {
+
+/// An unsupported model architecture was specified
+pub struct UnsupportedModelArchitecture(String);
```
philpax, line 112: I suggest implementing `std::error::Error` for this. Shouldn't need to implement any of its methods.

crates/llm/src/lib.rs
```diff
@@ -187,3 +199,26 @@ pub fn load_dynamic(
...
     Ok(model)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_model_architecture_from_str() {
+        static ARCHS: [ModelArchitecture; 5] = [
```
philpax, line 209: Maybe add this as an associated constant to `ModelArchitecture`, so that anyone can loop over `ModelArchitecture::ALL`?

crates/llm/examples/inference.rs
```diff
@@ -15,8 +15,7 @@ fn main() {
 
     let now = std::time::Instant::now();
 
-    let architecture = ModelArchitecture::from_tag(model_type)
-        .unwrap_or_else(|| panic!("{model_type} is not a supported model"));
+    let architecture = model_type.parse().unwrap_or_else(|e| panic!("{e}"));
```
philpax, line 18: Does the root-level `llm` example need to be changed to match this?

philpax (approved):

crates/llm/examples/inference.rs
danforbes: No, that one just uses LLaMA statically for simplicity.

PR #195, opened 2023-05-08
Rework the inference API

philpax:
Before this change, there were three structs with `Inference` and `Params` in their name. This fixes that by renaming them and consolidating the API.


hhamud (approved):
LGTM!

LLukas22:
Jep, better :D

PR #204, opened 2023-05-10
Implement RedPajama as a NeoX Specialization

danforbes:
Tested with https://huggingface.co/keldenl/RedPajama-INCITE-Chat-7B-v0.1-GGML/tree/main

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -89,7 +89,7 @@
           "kind": "example"
         }
       },
-      "args": ["gptneox-redpajama", "${env:HOME}/.ggml-models/redpajama-incite-7b.bin"],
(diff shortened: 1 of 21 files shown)


philpax:
Sorry about asking this again 😅 Are the changes between NeoX and RedPajama small enough to be covered by `use_parallel_residual`, or are there other changes? Because if it's just `use_parallel_residual`, we should figure out how to specify parameters for the model.

danforbes:
> Are the changes between NeoX and RedPajama small enough to be covered by `use_parallel_residual`

I think that may be possibly true, but I'm not sure if I understand how to use that intuition to create a better implementation.

philpax:
My changes move `use_parallel_residual` to the model's hyperparameters - now the problem to figure out is how to actually set those hyperparameters before/during load. (Changing them manually confirms that both implementations work.)

Will sleep on it 💤

Issue #205, opened 2023-05-10
How can I stop the inference to include the prompt tokens?


juliooa:
If I call this:
```
> llm llama infer -m <model_path> -p "Tell me how cool the Rust programming language is: "
```
I get:
```
✓ Loaded 291 tensors (4.2 GB) after 144ms
Tell me how cool the Rust programming language is:
The first thing that comes to mind when I think about rust... 
```
I'm not sure if this is something from `llm`, but it always includes the initial prompt in the inference, can I tell it to not do that?
When I try the same in llama.cpp it doesn't include the initial prompt.
To be clear, I want to achieve this:
```
✓ Loaded 291 tensors (4.2 GB) after 144ms
The first thing that comes to mind when I think about rust... 
```
I can manually take it out, but I'm curious if there is a way to remove it with the crate.

Thanks for the crate!

danforbes:
At this time that is not possible - prompt tokens are sent to the inference callback function just like all other tokens (Ref: https://github.com/rustformers/llm/blob/main/crates/llm-base/src/inference_session.rs#L100) There have been discussions about giving downstream users more information and control about how and why the callback function is being called.

philpax:
Copying this across from Discord: I like that proposal, but I'm not sure about having the error in the `Response` (I can't think of too many applications where you'd want to still run the callback after a failure), and the `Feedback` should have a way to return an error (`Halt` could potentially take a `Option<E: Error>`, maybe?)

Something like
```rust
pub enum InferenceResponse<'a> {
    PromptToken(&'a str),
    InferredToken(&'a str),
    EotToken,
}

pub enum InferenceFeedback<'a, E: Error + 'static> {
    Continue,
    FeedPrompt(&'a str),
    Halt(Option<E>),
}
```
perhaps?

juliooa:
Cool, thanks for the replies, this looks great.

danforbes:
Closing now that the above has been merged in.

philpax:
Not quite yet - there's still no way to hide the prompt being shown in the CLI. Fixing now...

danforbes:
Ah, I see...

philpax:
*Now* we're sorted 😂

Issue #151, opened 2023-04-23
Related to stop prompt 


hlhr202:
Hi, some models got very stupid stopping logic (eg. vicuna under instruct mode)
llama-rs can provide a way (like a stop prompt argument) to stop it by giving a text sequence.
What I v investigated are as follows:

https://github.com/Atome-FE/llama-node/blob/main/packages/llama-cpp/src/llama.rs#L152
https://github.com/sobelio/llm-chain/blob/main/llm-chain-llama/src/executor.rs#L96

I think llama.cpp also provide something called reverse prompt or anti prompt (but they just use it in interactive mode)

LLukas22:
I agree that adding stop word/sequence support natively would be helpful for using models in a chatbot setting, especially ones that aren't specifically trained for that purpose. Perhaps we could add a list of strings to the `InferenceParameters`, tokenize them, and then match the last N-generated tokens against the tokenized stop words. This should be easy enough and would prevent models from talking to themselves 😓.

danforbes:
Here is a naive implementation https://github.com/danforbes/llama-rs/commit/4d15b0995c471402fac5150dde4262c241e9902a

philpax:
@danforbes Sorry, keep forgetting to get back to you on this! Yeah that seems fine to me, want to add `antiprompt: Option<&str>` to `inference_with_prompt`?

danforbes:
You need to provide more information than just the "reverse prompt" - among other things you need to provide some kind of callback for receiving user input. There are also some changes to the way that returned tokens are handled (EOT handling as well as reverse prompt handling) that I wasn't sure how to implement cleanly in the existing `inference_with_prompt` function.

danforbes:
@hlhr202 this should be implemented with #206, but please open another Issue if you have more suggestions 🙏🏻

PR #206, opened 2023-05-10
Fine-grain inference feedback

danforbes:
Related to #151 and #205

### DIFF
--- a/crates/llm-base/src/util.rs
+++ b/crates/llm-base/src/util.rs
@@ -20,6 +20,8 @@ macro_rules! mulf {
 use memmap2::{Mmap, MmapAsRawDesc, MmapOptions};
 use thiserror::Error;
 
+use crate::inference_session::{InferenceFeedback, InferenceResponse};
+
 /// Used to buffer incoming tokens until they produce a valid string of UTF-8 text.
 ///
 /// Tokens are *not* valid UTF-8 by themselves. However, the LLM will produce valid UTF-8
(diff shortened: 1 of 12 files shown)


danforbes:
Just realized I never actually implemented `Halt` on the `InferenceSession` side 😅

danforbes:
I'm not able to see why the job is failing
image

philpax:
This looks good to me from a glance, but I'll need to pull it down locally and play around with it. Will try to get to it soon 🚀

crates/llm-base/src/util.rs
```diff
@@ -60,14 +62,15 @@ impl TokenUtf8Buffer {
...
     pub fn adapt_callback<'a, E: std::error::Error + 'static>(
-        mut callback: impl FnMut(&str) -> Result<(), E> + 'a,
-    ) -> impl FnMut(&[u8]) -> Result<(), E> + 'a {
+        mut callback: impl FnMut(InferenceResponse) -> Result<InferenceFeedback, E> + 'a,
+    ) -> impl FnMut(&[u8]) -> Result<crate::inference_session::InferenceFeedback, E> + 'a {
         let mut buffer = Self::new();
         move |token| match buffer.push(token) {
-            Some(tokens) => callback(&tokens),
-            None => Ok(()),
+            Some(tokens) => callback(InferenceResponse::PromptToken(tokens)),
```
philpax, line 72: Hm, might be a little counterintuitive for it to always be a `PromptToken`. I'd suggest adding that to the name of the function (it seems like we only ever use it in a prompt context) or adding a parameter to switch between prompt and inference tokens.

philpax (approved):
LGTM, apart from that one quibble.

crates/llm-base/src/util.rs
danforbes: Okay, I just pulled it into `inference_session.rs` and gave it a more appropriate name :+1:

PR #217, opened 2023-05-12
MPT model support

LLukas22:
These are just the Hyperparameters and the CLI integration to to enable the use of the `quantize` function. 

Inferenze and actual model loading still needs to be implemented!

### DIFF
--- a/crates/models/mpt/src/lib.rs
+++ b/crates/models/mpt/src/lib.rs
@@ -82,6 +82,7 @@ impl KnownModel for Mpt {
 /// MPT [hyperparameters](https://en.wikipedia.org/wiki/Hyperparameter_(machine_learning))
 #[derive(Debug, Default, PartialEq, Eq, Clone, Copy)]
 pub struct Hyperparameters {
+    //for the love of god please rename these to something more descriptive and standardise the names accross the different models 
     /// TODO: document
     d_model: usize,
     /// TODO: document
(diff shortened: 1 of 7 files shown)


philpax:
A start's a start! fyi @danforbes

LLukas22:
I will upload the quantized models to this repo: https://huggingface.co/LLukas22/mpt-7b-ggml (Uploading will take some time)

danforbes:
Looks like they're available?

crates/models/mpt/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+    fn inference_parameters(&self) -> &InferenceParameters {
+        &self.inference_params
+    }
+
+}
+
+/// MPT [hyperparameters](https://en.wikipedia.org/wiki/Hyperparameter_(machine_learning))
+#[derive(Debug, Default, PartialEq, Eq, Clone, Copy)]
+pub struct Hyperparameters {
+    //for the love of god please rename these to something more descriptive and standardise the names accross the different models
```
danforbes, line 85: Yeah, the names leave a lot to be desired, but for now I have taken the approach of trying to mimic the nomenclature used in the "reference" C++ implementation in order to help make debugging easier for myself. Another thing is that in many (most) cases I don't actually know the purpose of the data...I've tried to add documentation to the ones that I feel comfortable doing so for.

philpax: Yeah, we should definitely go through and fix these up. I think a lot of the original names can be recovered from looking at the Python / HF model docs.

danforbes:
Closing in favor of #218

PR #232, opened 2023-05-16
Added support for multiple LoRA adapters

### LINKED ISSUE #227: Support multiple LoRA adapters
Accoring to @NatanFreeman `peft` supports the use of multiple LoRA adapters for the same model. 
The following is valid python code:
```python
from peft import PeftModel
from transformers import LLaMATokenizer, LLaMAForCausalLM, GenerationConfig

tokenizer = LLaMATokenizer.from_pretrained("decapoda-research/llama-7b-hf")

model = LLaMAForCausalLM.from_pretrained(
    "decapoda-research/llama-7b-hf",
    load_in_8bit=True,
    device_map="auto",
)
model = PeftModel.from_pretrained(model, "Naataan/Model-A")
model = PeftModel.from_pretrained(model, "Naataan/Model-B")

# save the applied weights to a new bin file
PROMPT = """Below is an instruction that describes a task. Write a response that appropriately completes the request.
### Instruction:
Natan: Hello.
Name the 3 biggest cities in England
### Response:"""

inputs = tokenizer(
    PROMPT,
    return_tensors="pt",
)
input_ids = inputs["input_ids"].cuda()

generation_config = GenerationConfig(
    temperature=0.6,
    top_p=0.95,
    repetition_penalty=1.15,
)
print("Generating...")
generation_output = model.generate(
    input_ids=input_ids,
    generation_config=generation_config,
    return_dict_in_generate=True,
    output_scores=True,
    max_new_tokens=128,
)
for s in generation_output.sequences:
    print(tokenizer.decode(s))
```

We probably have to check how `peft` handles the patching of multiple tensors and implement it.

LLukas22:
Closes #227.

Moved the loading logic into the `LoraPatches` struct  and added the `LoraAdapter` struct which abstracts multiple LoRA patches.

Changed "loar_adapter" to "lora_adapters" in the `ModelParameters`.


LLukas22:
Alright i will move the loading logic back over and just iterate over the LoraAdapter's while loading the tensors.

LLukas22:
This should be ready, but i havent tested it yet.

LLukas22:
Old LoRA adapters fail with the following error:
```
✗ Failed to load model    
Error:
   0: Could not load model
   1: invariant broken: 256 <= 2 in Some("E:\\GGML_Models\\LoRA\\alpaca-7B.bin")
```
Is this a side-effect of the new quantization formats?

philpax:
Can you provide more information about which models you tried it with? That sounds like something's misaligned because it comes from https://github.com/rustformers/llm/blob/923a09bf27f1fedf366d33ca2e4359c1eba8cd8a/crates/ggml/src/format/loader.rs#L209

PR #237, opened 2023-05-17
Update Gpt-NeoX implementation

LLukas22:
This adds the new boolean `use_parallel_residual` hyperparameter to the Gpt-Neox implementation.

This also enables us to remove the RedPajama model overwrite.

I will use this branch to convert and quantize the models published by [Together](https://huggingface.co/togethercomputer).

### DIFF
--- a/binaries/llm-cli/src/cli_args.rs
+++ b/binaries/llm-cli/src/cli_args.rs
@@ -37,21 +37,6 @@ pub enum Args {
     NeoX {
         #[command(subcommand)]
         args: BaseArgs,
-
-        #[arg(long)]
-        /// By default, the GPT-NeoX architecture uses a parallel residual.
-        ///
-        /// This flag disables that, as some models out there are trained without it,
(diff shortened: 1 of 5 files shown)


LLukas22:
Tested with these models: https://huggingface.co/LLukas22/redpajama-ggml

The output sometimes contains the padding Token, which should not be the case and the first generated tokens seem to be gibberish.

philpax:
Can't test `use_parallel_residual=false` right now because I don't have a RedPajama model at hand, but looks pretty promising otherwise.

Had a bit of a look at why #246 is happening. It looks like [the upstream implementation](https://github.com/ggerganov/ggml/blob/master/examples/gpt-neox/main.cpp) has changed a fair bit, and makes fewer changes for the `use_parallel_residual` case. I think we'll need someone to go through and reconcile the two again.

philpax:
I'm going to merge this, because the issue isn't caused by this PR, and we'll need to fix it anyway.

PR #247, opened 2023-05-18
Typo fix in extern function name

tanmaysachan:
Was probably being optimized away as dead code, hence no error.


philpax:
Thanks! This was also fixed in #217, but this fix shouldn't conflict.

PR #248, opened 2023-05-19
Add perplexity calculation and cli-arg

tanmaysachan:
Borrowed the implementation from https://huggingface.co/docs/transformers/perplexity

llama.cpp implements it a little differently, and their implementation does not work on prompts of size less than the context window.

For the interface, I added a simple cli-arg to run a perplexity check.

Solves https://github.com/rustformers/llm/issues/90.

### DIFF
--- a/crates/llm-base/src/inference_session.rs
+++ b/crates/llm-base/src/inference_session.rs
@@ -5,7 +5,8 @@ use rand::{distributions::WeightedIndex, prelude::Distribution};
 use thiserror::Error;
 
 use crate::{
-    mulf, InferenceError, InferenceParameters, Model, OutputRequest, TokenId, TokenUtf8Buffer,
+    mulf, util::softmax, InferenceError, InferenceParameters, Model, OutputRequest, TokenId,
+    TokenUtf8Buffer,
 };
 
 // The size of a scratch buffer used for inference. This is used for temporary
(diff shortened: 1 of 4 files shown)


crates/llm-base/src/inference_session.rs
```diff
@@ -191,7 +192,9 @@ impl InferenceSession {
         )?;
         stats.feed_prompt_duration = start_at.elapsed().unwrap();
         stats.prompt_tokens = self.n_past;
-
+        if request.run_perplexity {
+            self.perplexity(model, parameters, request, output_request);
```
philpax, line 196: Consider adding a `perplexity: Option<f32>` to the `stats` (since they're available here), and setting it if `run_perplexity` is specified. You can then use these stats for printing in the CLI.

philpax:
Looks good to me outside of those two comments - great work, thanks for contributing!

crates/llm-base/src/inference_session.rs
```diff
@@ -225,6 +228,67 @@ impl InferenceSession {
...
+    pub fn perplexity(
+        &mut self,
+        model: &dyn Model,
+        parameters: &InferenceParameters,
+        request: &InferenceRequest,
+        output_request: &mut OutputRequest,
+    ) {
+        // Setup the output request to store logits.
+        output_request.all_logits = Some(Vec::new());
+        let mut prompt_tokens = model
```
philpax, line 241: Hm, is this the only use of `request`? It could be worth taking a `&[TokenId]` instead and letting users decide if they want to use `InferenceSession::tokens` or pass in their own tokens (potentially with a hint about tokenization)

tanmaysachan: Like adding a cli for passing tokens to `--perplexity` that are different from `-p`?

philpax: Ah, don't worry about it for now, this is fine :)

tanmaysachan:
@philpax I think there are a few errors with the calculation right now for large text files. Could you revert this PR for now if possible? there's a few possible panic cases.

PR #255, opened 2023-05-20
Model architectures should appear in alphabetic order

danforbes:


philpax (approved):
Yeah sure why not

danforbes:
> Yeah sure why not

I get triggered when they don't 😅

PR #256, opened 2023-05-20
More model standardization

danforbes:


philpax:
Looks reasonable, are there any changes to the behaviour at all?

danforbes:
No changes in behavior...at least not intentional 😅 This is the only change to the public API that I'm aware of https://github.com/rustformers/llm/pull/256/files#diff-a97d4cc4f56e414e997a3fbd502b1875578e052e3fee62e8f363ac74cb5f56a4R159 It would be nice if you could take a quick look at the changes to variable names in the model definitions and evaluation functions just to make sure I have not introduced any misleading names (I tried to change names to make things _more_ clear).

philpax (approved):
One day we'll have automated tests for each of these models. I wonder how much GitHub will get mad at us for downloading gigabytes of models?

PR #262, opened 2023-05-21
Remove LLaMA "old_loader"

danforbes:


philpax:
Yeah, I suppose multipart models are rare enough these days that we don't need to keep this around.

PR #263, opened 2023-05-21
feat: cloning trait (DRAFT)

### LINKED ISSUE #240: [RFC] Ability to clone a Model
Right now when creating a model reference using load_dynamic, it returns a Box<dyn Model>. When running 2 inference session in parallel, this proves to be inefficient as the 2 instance will be racing to read from the same model heap.

The solution to this is to clone the model (clone the whole heap), so that each inference session have a dedicated model reference to read from, avoid the whole read competition.

It's not possible to clone at the moment since the reference returned from load_dynamic is a `Box<dyn Model>` which upon clone() would only make a copy of the reference, not the whole heap itself.

The solution is to implement a `clone_box` function, OR the ability to do `Model::from` which copy the full heap into a clone, which can be used for new inference session.

louisgv:
closes #240


crates/llm-base/src/model/mod.rs
```diff
@@ -205,8 +205,18 @@ pub trait Model: Send + Sync {
     /// [InferenceSession::infer]). This value is configured through
     /// [ModelParameters::inference_parameters].
     fn inference_parameters(&self) -> &InferenceParameters;
+
+    /// Clone this model into a boxed trait object.
+    fn clone_box(&self) -> Box<dyn Model>;
```
philpax, line 210: Incredibly minor: I prefer `clone_boxed` just to indicate that the clone is being boxed, not that the box itself is being cloned

crates/models/bloom/src/lib.rs
```diff
@@ -45,6 +45,15 @@ pub struct Bloom {
 
 unsafe impl Send for Bloom {}
 unsafe impl Sync for Bloom {}
+impl Clone for Bloom {
+    fn clone(&self) -> Bloom {
+        // Clone the fields here
+        Bloom {
+            // field1: self.field1,
```
philpax, line 52: Yeah, this was about as far as I got when I started implementing it myself 😅 

The tensors need to be cloned into the new context. We should define a method for cloning a tensor from another context.

philpax:
I'm going to close this because it's out of date now, but we should revisit this at some point!

PR #270, opened 2023-05-23
Pull upstream changes into GptNeox 

### LINKED ISSUE #246: Update to latest upstream GPT-NeoX implementation / Fix GPT-NeoX inference
GPT-NeoX inference seems a little... funky, as reported by @LLukas22:

```
1: <|padding|>The meaning of life is the name.
“I was doing this for a while, and I knew it could make an impact,” says Coker about his time in college football coaching roles at North Carolina State University (NCSU) from 2004 to 2013 before going on to be head coach – again! -of NEC Conference’s Delaware Division of Basketball.
But then came the 2016-17 season, when he became a first year assistant for UNC women's basketball under Hall Of Fame Head Coach Sylvia Hatchell where Coker helped in leading her team through an historic NCAA National Championship run that saw them defeat Notre Dame

2: <|padding|>The meaning of life is the only way to be, or not at least a partialwayto-ofex: part_LATEX LAC PARTICIELPARTISTCORLECRECLPINORTHCAREBUT)
(Photo by Alex Wong/Getty Images), (Source: AP Photo / Evan Vucci)(AP Photos): President Barack Obama speaks about the proposed Keystone XL pipeline in Washington. A judge has given TransCanada Corp permission to build a controversial oil sands project that would cross an international border and run through US-owned land — but it was not clear Monday how long he will keep his word.(Photo by Evan Vucci/AP)
(Source: AP Photo / John Minchillo)(Reuters): Democratic presidential candidate Hillary Clinton holds up the key to a new office in her New York home. As she ha

3.<|padding|>The meaning of life is a short-post, post by an easy to use and detailed?PostByGetAButContinueEcheLIfYouTWhatIsIt?HowToBeginAfterGAtthes the details are too long for you (or just)
I’m not going anywhere. I am here forever; so it seems as if there is no end in sight—that life has run its course and that we will be back again at a later time to finish what was started, but our destiny? Maybe this story isn't over yet: After all the world keeps turning
```

This was captured with a RedPajama model.

LLukas22:
Fixes #246.

The [tokenization behaviour](https://github.com/rustformers/llm/blob/main/crates/llm-base/src/vocabulary.rs#L123-L126) has to be changed to make this work with every gpt-neox based model.
e.g. Redpajama doesn't like the added padding token at the start of a sequence.

### DIFF
--- a/crates/models/gptneox/src/lib.rs
+++ b/crates/models/gptneox/src/lib.rs
@@ -6,7 +6,7 @@ use std::error::Error;
 
 use ggml::Tensor;
 use llm_base::{
-    ggml::{self},
+    ggml,
     model::{common, HyperparametersWriteError},
     util, FileType, InferenceParameters, InferenceSession, InferenceSessionConfig, KnownModel,
     LoadError, Mmap, ModelParameters, OutputRequest, TensorLoader, TokenId, Vocabulary,
@@ -41,34 +41,6 @@ pub struct GptNeoX {
     // must be kept alive for the model
     _context: ggml::Context,
     _mmap: Option<Mmap>,
-}
-
-// feed-forward network
-fn _gpt_neox_ff(layer: &Layer, context: &ggml::Context, input: &Tensor) -> Tensor {
-    let mut current = context.op_norm(input);
-
-    //gain and bias
-    current = context.op_add(
-        &context.op_mul(&context.op_repeat(&layer.ln_2_g, &current), &current),
-        &context.op_repeat(&layer.ln_2_b, &current),
-    );
-
-    //apply weights
-    current = context.op_mul_mat(&layer.c_mlp_fc_w, &current);
-
-    //apply bias
-    current = context.op_add(&context.op_repeat(&layer.c_mlp_fc_b, &current), &current);
-
-    // GELU activation
-    current = context.op_gelu(&current);
-
-    // projection
-    // cur = proj_w*cur + proj_b
-    current = context.op_mul_mat(&layer.c_mlp_proj_w, &current);
-
-    current = context.op_add(&context.op_repeat(&layer.c_mlp_proj_b, &current), &current);
-
-    current
 }
 
 unsafe impl Send for GptNeoX {}
@@ -162,7 +134,7 @@ impl KnownModel for GptNeoX {
         )
     }
 
-    //allow snake case here as its a one-to-one mapping of the original names
+    // allow snake case here as its a one-to-one mapping of the original names
     #[allow(non_snake_case)]
     fn evaluate(
         &self,
@@ -328,7 +300,7 @@ impl KnownModel for GptNeoX {
             let feedforward_input: Tensor;
             if !use_parallel_residual {
                 feedforward_input = ctx0.op_add(&current, &input_layer);
-                current = _gpt_neox_ff(&self.layers[il], &ctx0, &feedforward_input);
+                current = feed_forward_network(&ctx0, &self.layers[il], &feedforward_input);
                 // input for next layer
                 input_layer = ctx0.op_add(&current, &feedforward_input);
             } else {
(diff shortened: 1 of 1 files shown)


crates/models/gptneox/src/lib.rs
```diff
@@ -6,7 +6,7 @@ use std::error::Error;
 
 use ggml::Tensor;
 use llm_base::{
-    ggml::{self, ElementType},
+    ggml::{self},
```
philpax, line 9: replace with `ggml`

crates/models/gptneox/src/lib.rs
```diff
@@ -43,6 +43,34 @@ pub struct GptNeoX {
     _mmap: Option<Mmap>,
 }
 
+// feed-forward network
+fn _gpt_neox_ff(layer: &Layer, context: &ggml::Context, input: &Tensor) -> Tensor {
```
philpax, line 47: Remove underscore, move to the end of the file (the public-facing stuff should come first)

crates/models/gptneox/src/lib.rs
```diff
@@ -43,6 +43,34 @@ pub struct GptNeoX {
...
+    let mut current = context.op_norm(input);
+
+    //gain and bias
```
philpax, line 50: Space after the `//`

crates/models/gptneox/src/lib.rs
```diff
@@ -182,95 +216,55 @@ impl KnownModel for GptNeoX {
...
 
-            gf.build_forward_expand(&ctx0.op_cpy(&kcur, &little_k));
-            gf.build_forward_expand(&ctx0.op_cpy(&vcur, &little_v));
+            gf.build_forward_expand(&ctx0.op_cpy(&kcur, &k));
+            gf.build_forward_expand(&ctx0.op_cpy(&vcur, &v));
 
-            let q = ctx0.op_permute(&qcur, 0, 2, 1, 3);
-            let big_k = ctx0.op_permute(
+            // Q = Qcur.contiguous().view(n_embd/n_head, n_head, N).permute(0, 2, 1, 3)
+            let Q = ctx0.op_permute(&qcur, 0, 2, 1, 3);
```
philpax, line 265: I'd suggest still lowercasing the names, if only because Rust syntax highlighting assumes that anything uppercase is a constant

philpax (approved):
Great work! Minor style nits but apart from that this is good to go from my end.

Regarding the BOS: does upstream have the same issue?

LLukas22:
Nope upstream is not affected by the BOS token schenanigans. Seams like they don't append anything at the beginning.

crates/models/gptneox/src/lib.rs
LLukas22: Well i actually don't know about this, at some point in the future we should/have to replace this with some attention head abstractions until then keeping it in sync with upstream is easier, if we keep the same names.

philpax:
Maybe we should just set bos=false for GPT-NeoX? That feels like a weird hack, but I can't see anything in the HF config that controls this.

@Narsil sorry about summoning you here, but for a given HF model, how do we know if the BOS token should be added / what I presume is `add_special_tokens` [here](https://docs.rs/tokenizers/latest/tokenizers/tokenizer/struct.Tokenizer.html#method.encode) (we're switching to HF in #271) should be turned on? Does it have to be bounced up to the user?

crates/models/gptneox/src/lib.rs
philpax: Agreed, but will downcasing cause any issues? It should be pretty obvious they're the same thing... but if there's at all any conflicts then yeah leave it as-is, I'm not that picky about this

Narsil:
@philpax No worries for the summon, happy to help.

`add_special_tokens` should be True by default yes.

But I think also what you are looking for is `post_process`.

The `post_processor` in `tokenizers` is the method that takes a vec of encoded strings (usually a single or a pair of tokenized inputs, like question and context for question answering)
This merges them into a single sequence to be consumed by the model:

Here is an example
https://github.com/Narsil/smelte-rs/blob/main/examples/gpt2.rs#L261-L262

Unfortunately, some tokenizers are not necessarily correctly configured on the hub, because some special tokens are sent directly within the string itself (especially useful for when it's a multilingual model like Whisper or a translation model). So some knowledge might remain out of the tokenizer itself.

`encode + post_process` should be the go-to I think.

PR #275, opened 2023-05-24
Update GGML Dependency & Bindings

danforbes:


philpax:
I've extended the bodge in the loader to work around https://github.com/ggerganov/llama.cpp/issues/1590.

Apart from that, this seems to work fine for me with https://huggingface.co/TheBloke/koala-7B-GGML/blob/main/koala-7B.ggmlv3.q5_1.bin . Haven't tested the other models yet because I don't have fresh variants of them on hand, but I don't see any reason for them to be broken.

PR #276, opened 2023-05-25
Add interface to allow changing of logits and sampling before inference

tanmaysachan:
Can something along these lines address https://github.com/rustformers/llm/issues/267?
`customize_logits` can be passed `Option<&Fn>` without the bits from the model to control logit distribution and sampling strategy.

The `TokenId` is returned as an option to further not require passing around of `params` and `rng`.

Can add interface for params to `customize_logits` in an edit if this implementation is enough.

### DIFF
--- a/crates/llm-base/src/inference_session.rs
+++ b/crates/llm-base/src/inference_session.rs
@@ -123,8 +123,22 @@ impl InferenceSession {
             return Err(InferenceError::ContextFull);
         }
 
-        // First, sample the next token, using the stored last_logits;
-        let next_token = self.sample_top_p_top_k(params, rng);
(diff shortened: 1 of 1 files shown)


philpax:
Hey there! Thanks for submitting the PR, and sorry for not getting around to it earlier - I actually needed it for some deterministic testing stuff that I wanted to do, but realized that when looking at it that we could evolve this to support entirely custom samplers.

I ended up doing that work in #280 because I changed how all of the relevant code works, but thank you for the inspiration 🙂 

I'm going to close this PR in favour of that one, but thanks for the contribution once again! If you're interested/not already there, you may also want to join [our Discord](https://discord.gg/YB9WaXYAWU).

PR #278, opened 2023-05-25
Add scratchbuffers to MPT

LLukas22:
Add scratchbuffers similar to the upstream ggml examples of GptNeoX/MPT. Should reduce memory consumption.

### DIFF
--- a/crates/ggml/src/tests.rs
+++ b/crates/ggml/src/tests.rs
@@ -46,7 +46,7 @@ fn will_fail_on_scored_ggml_save() {
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


crates/ggml/src/tests.rs
```diff
@@ -46,15 +46,15 @@ fn will_fail_on_scored_ggml_save() {
 }
 
 #[test]
-fn can_roundtrip_loader_and_saver_ggjt_v2() {
+fn can_roundtrip_loader_and_saver_ggjt_v3() {
```
philpax, line 49: oops good catch

Issue #274, opened 2023-05-24
Context clone may not actually preserve the memory


philpax:
Need to test this, but

https://github.com/rustformers/llm/blob/c51ff5d60b4177d2c75ee8483c25b2b4ee88f07e/crates/llm-base/src/inference_session.rs#L501-L520

does not copy the memory from the existing session, which seems suspect. We probably want a function to copy a tensor from one to another - #263 will benefit from this too

philpax:
Ok, we don't actually support `Clone` on `InferenceSession` any more. We should fix that, but in the meantime, I'll drop this from the milestone.

philpax:
Thought about this some more. We should have `Clone` in the session - it would be a regression to ship without it :(

PR #280, opened 2023-05-27
Sampler rework

philpax:
This change follows on from #276; I needed a way to do deterministic sampling and realised that it wouldn't be too hard to introduce a trait to let you have full control over sampling behaviour.

The only thing I'm unhappy about is the use of `Arc<dyn Sampler>` - I would have preferred taking a reference in `InferenceParameters`, but that would make implementing `Default` impossible. I decided to compromise and take an `Arc` instead, but, you know, eh.


LLukas22:
Shouldn't the sampler be in the inference session? I don't think adding it to the `InferenceParameters ` is the right call. Maybe we could pass it into the session constructor?

philpax:
That's actually what I thought and did the first time, but I noticed a few things:
1) the existing interface lets you vary the sampling parameters, so we would not be able to replicate the existing interface's functionality if you did that
2) because the sampler is type-erased, you can't `Serialize` it, which is required for `InferenceSessionParameters` and `InferenceSnapshot`

LLukas22:
In my oppinion the session is getting a bit bloated. And i dont think we need to serialize the sampler, we just need to serialize the k,v memory and the last tokens. The user should be able to pass any sampler he wants into the session.

philpax:
> In my oppinion the session is getting a bit bloated. 

Hm, not sure I agree - I can definitely see the argument that at a glance it carries more than it needs to, but I can't see anything that should be eliminated that won't reduce its usefulness.

> And i dont think we need to serialize the sampler, we just need to serialize the k,v memory and the last tokens. 

That's possible, but what would it look like?

```rust
let snapshot = unsafe { session.get_snapshot() };
let new = InferenceSession::from_snapshot(snapshot, model);
new.set_sampler(samplers::TopPTopK { ... });
```

It would be a bit odd to me if you had to set the sampler on a recreated session to do anything with it.

Also, what if I'm testing out different sampling strategies at runtime, or doing something we haven't accounted for, like a sampler that changes behaviour at runtime?

I went for this design because - even if there was a way to preserve the sampler fully (which I think could be done with a bit of hackery) - this is still more flexible and behaves similarly to the existing interface.

LLukas22:
Understood, I see your point. I might not be using the snapshot feature enough to provide valuable feedback. Mostly, I just create a session, generate some text, and then end the session.

But you're right, setting the sampler externally does seem a bit odd.😅

philpax:
Yeah, totally fair! To clarify what I wanted to do with this - I wanted to test / fix #274 by dropping in a sampler that always samples the most-probable token, then cloning the context after feeding the prompt and comparing the inference output between the two. I wanted to be able to use the same most-probable sampler for both inference attempts.

I'm going to see if I can get rid of that `Arc`, and then merge this.

philpax:
So there's a fairly controversial change here - the removal of `InferenceParameters::default`. This is because it takes a reference to the `Sampler`, which can't be defaulted.

This is something that I arrived at after a fair bit of pondering:

- We could have a `Box<dyn Sampler>` that gets defaulted, but then you can't clone it
- We could have an `Arc<dyn Sampler>`, but then you're reference counting something that doesn't need to be reference counted, and supplying a custom sampler will require allocation, especially if you want to swap it out between inferences
- We could have an `Option<&dyn Sampler>` and then use Top-P Top-K by default in the inference code, but that obscures the actual default settings - the settings would be resolved in `InferenceSession`, away from where the default would be defined
- We could have a generic on `InferenceSession` for the sampler, but this complicates things immensely because now you have a generic floating around that needs to be resolved
- We could use `Cow<Sampler>`, but there is no easy way to define a `ToOwned<Owned = Box<dyn Sampler>>` for all `S: Sampler` due to trait coherence issues

Ideally, the `InferenceParameters` should:
- be defaultable
- be cheap to copy/clone
- be readily reusable across inference calls
- be easy to use, without having to know too much about Rust intricacies
- be explainable/obvious - no magic at a distance

I'm still not sure what the best play would be here.

philpax:
Yeah, you're right, overthinking this. I'll go back to `Arc` :)

PR #290, opened 2023-06-01
Fix error in Usage section of README

danforbes:


LLukas22:
Could you extend the examples to include the optional HF Tokenizer?

LLukas22:
LGTM 👍

Issue #288, opened 2023-06-01
[Bug] Dimension mismatch when using crates/llm/examples/embeddings.rs


Jeadie:
I'm attempting to run the basic embedding example [embeddings.rs](https://github.com/rustformers/llm/blob/main/crates/llm/examples/embeddings.rs). I am getting the following error (RUST_BACKTRACE=1).
```shell
Loaded hyperparameters
ggml ctx size = 3567.88 MB

Loaded tensor 8/194
Loaded tensor 16/194
Loaded tensor 24/194
Loaded tensor 32/194
Loaded tensor 40/194
Loaded tensor 48/194
Loaded tensor 56/194
Loaded tensor 64/194
Loaded tensor 72/194
Loaded tensor 80/194
Loaded tensor 88/194
Loaded tensor 96/194
Loaded tensor 104/194
Loaded tensor 112/194
Loaded tensor 120/194
Loaded tensor 128/194
Loaded tensor 136/194
Loaded tensor 144/194
Loaded tensor 152/194
Loaded tensor 160/194
Loaded tensor 168/194
Loaded tensor 176/194
Loaded tensor 184/194
Loaded tensor 192/194
Loading of model complete
Model size = 3568.33 MB / num tensors = 194

thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `7`,
 right: `28672`', /Users/jeadie/.cargo/git/checkouts/llm-d8a8bbe144aa0546/b3c9851/crates/llm-base/src/model/common.rs:88:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/std/src/panicking.rs:579:5
   1: core::panicking::panic_fmt
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/panicking.rs:64:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/panicking.rs:211:5
   4: llm_base::model::common::extract_embeddings
             at /Users/jeadie/.cargo/git/checkouts/llm-d8a8bbe144aa0546/b3c9851/crates/llm-base/src/model/common.rs:88:9
   5: <llm_mpt::Mpt as llm_base::model::KnownModel>::evaluate
[...]

PR #291, opened 2023-06-02
Fix embeddings extraction for all models

skirodev:
Apply the fixed code for embeddings extraction to all models to avoid assertion errors. #288

### DIFF
--- a/crates/llm/examples/embeddings.rs
+++ b/crates/llm/examples/embeddings.rs
@@ -49,11 +49,7 @@ fn main() {
     };
 
     // Load model
-    let model_params = llm::ModelParameters {
-        prefer_mmap: true,
-        context_size: 4096,
-        lora_adapters: None,
-    };
+    let model_params = llm::ModelParameters::default();
     let model = llm::load_dynamic(
         model_architecture,
         &model_path,
@@ -124,7 +120,7 @@ fn get_embeddings(
     let vocab = model.vocabulary();
     let beginning_of_sentence = true;
     let query_token_ids = vocab
-        .tokenize(&format!(" {}", query), beginning_of_sentence)
+        .tokenize(&format!("{}", query), beginning_of_sentence)
         .unwrap()
         .iter()
         .map(|(_, tok)| *tok)
--- a/crates/models/bloom/src/lib.rs
+++ b/crates/models/bloom/src/lib.rs
@@ -344,6 +344,8 @@ impl KnownModel for Bloom {
             &input_layer,
         );
 
+        let embeddings_tensor: ggml::Tensor = input_layer.share();
+
         // lm_head
         input_layer = ctx0.op_mul_mat(&self.output, &input_layer);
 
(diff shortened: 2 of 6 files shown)


LLukas22:
Correct me if im wrong, until now we passed the output of the embedding layer as the model embedding?

philpax:
Hmm yeah, I think this is correct (the embeddings should come from just before the LM head is evaluated to coalesce the LM output into logits). I'm on holiday so I can't check, but I assume `embd` is the input embeddings, not the output.

Happy to merge if you can validate my understanding.

LLukas22:
In my oppinion this is also correct, but i'll test it when i get home. If it works i'll merge it.

skirodev:
@LLukas22 Yes, exactly, we obtain the vector corresponding to the last token as the embeddings for the whole sentence. The length of this embeddings is a fixed value equal to n_embd. For the LLaMA 7B model, its length is 4096, while for OpenAI's [`text-embedding-ada-002`](https://openai.com/blog/new-and-improved-embedding-model) model, its length is 1536.

LLukas22:
Good catch, when i get home from Work i'll test it against the rustformers HF models, but i'm 99% sure it will work.

Jeadie:
Confirmed that this is a fix for #288

LLukas22:
This works, but for gpt-j and gpt-neox based models im getting a `process didn't exit successfully: target\release\examples\embeddings.exe gptj C:\Users\Lu.Kreuss\Downloads\gpt-j-6b-q4_0-ggjt.bin -r EleutherAI/gpt-j-6b (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)` error. Tested with [rustformers/gpt-j-ggml](https://huggingface.co/rustformers/gpt-j-ggml/tree/main) and [rustformers/redpajama-ggml](https://huggingface.co/rustformers/redpajama-ggml).


Im also getting very poor embeddings this way, compared to some BERT models i have lying around. Maybe we should perform some sort of pooling on the embeddings of all tokens. The [SGPT](https://arxiv.org/pdf/2202.08904.pdf) paper uses weighted mean pooling, where more recent tokens have a stronger impact on the produced embedding than older tokens. Maybe this would improve the quality of the embeddings?

skirodev:
> This works, but for gpt-j and gpt-neox based models im getting a process didn't exit successfully: target\release\examples\embeddings.exe gptj C:\Users\Lu.Kreuss\Downloads\gpt-j-6b-q4_0-ggjt.bin -r EleutherAI/gpt-j-6b (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION) error. Tested with [rustformers/gpt-j-ggml](https://huggingface.co/rustformers/gpt-j-ggml/tree/main) and [rustformers/redpajama-ggml](https://huggingface.co/rustformers/redpajama-ggml).

This is because the value of "context_size" set in the model parameters was too large, so I rewrote the embeddings example.

> Im also getting very poor embeddings this way, compared to some BERT models i have lying around. Maybe we should perform some sort of pooling on the embeddings of all tokens. The [SGPT](https://arxiv.org/pdf/2202.08904.pdf) paper uses weighted mean pooling, where more recent tokens have a stronger impact on the produced embedding than older tokens. Maybe this would improve the quality of the embeddings?

This might be a viable approach, but I am not sure if it is necessary to use a model with a large number of parameters to generate embeddings.

LLukas22:
> This is because the value of "context_size" set in the model parameters was too large, so I rewrote the embeddings example.

o_O I feel kinda stupid for missing this, good job now everything works as expected 👍 

> This might be a viable approach, but I am not sure if it is necessary to use a model with a large number of parameters to generate embeddings?

It makes sense if you don't want to load a second smaller model to perform the embedding task. But i'm gooing to create another issue for this. It's not part of this PR.

PR #292, opened 2023-06-02
Made Sampler and InferenceParameters threadsafe

LLukas22:


danforbes (approved):
Looks very simple and was informally approved by @philpax in Discord, so I feel safe merging.

crates/llm-base/src/lib.rs
```diff
@@ -85,6 +85,11 @@ pub struct InferenceParameters {
     /// sampler that offers a [Default](samplers::TopPTopK::default) implementation.
     pub sampler: Arc<dyn Sampler>,
 }
+
+//Since Sampler implements Send and Sync, InferenceParameters should too.
+unsafe impl Send for InferenceParameters {}
```
philpax, line 90: hol up is this necessary? won't these be inherited?

philpax: (it's fine but I don't think it's needed?)

LLukas22: PyO3 was complaining, so i had to wrap it in an Arc<>. With this i can just use it in a python objekt.

philpax: weird, I think it should automatically infer Send + Sync for inference parameters as all the fields should be, too 🤔 

It's totally fine as is, but can you try playing around with it and see if it still compiles without the unsafe impls?

PR #296, opened 2023-06-05
Default to f16 model memory k/v in llm CLI and InferenceSessionConfig

KerfuffleV2:
Using 32bit values doesn't increase quality in a measurable way, see discussion here: https://github.com/ggerganov/llama.cpp/discussions/1593

The current default doubles memory consumption/size of prompt caches without a measurable upside. This pull changes the default to 16bit.

Allows `--float16` for backward compatibility.

Adds `--no-float16` flag to allow using 32bit memory.

***

Don't know if it's just me but I can't even run `llm llama infer --help` from the main branch:

```
Message:  Command infer: Argument names must be unique, but 'vocabulary_path' is in use by more than one argument or group
```

### DIFF
--- a/binaries/llm-cli/src/cli_args.rs
+++ b/binaries/llm-cli/src/cli_args.rs
@@ -244,10 +244,16 @@ pub struct Generate {
     #[arg(long, default_value = None)]
     pub seed: Option<u64>,
 
-    /// Use 16-bit floats for model memory key and value. Ignored when restoring
-    /// from the cache.
-    #[arg(long, default_value_t = false)]
-    pub float16: bool,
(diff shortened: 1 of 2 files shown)


KerfuffleV2:
@LLukas22 Thanks! I missed that one.

> Regarding the --help parameter the application only panics in debug mode

It's really weird that compiling in release affects argument parsing. I'd never have guessed to try a different build type.

LLukas22:
Yeah `clap` seams to validate duplicated parameters only in debug mode. 🤷 

Thanks for changing the defaults, everything looks good now. 👍

PR #300, opened 2023-06-07
Fix `--help` panicing in debug mode

### LINKED ISSUE #297: `--help` Parameter panics in debug mode
As mentioned in https://github.com/rustformers/llm/pull/296 the `--help` parameter panics with:
```
Message:  Command infer: Argument names must be unique, but 'vocabulary_path' is in use by more than one argument or group
```
When run in debug mode.

When running in release mode the aplication does not panic.

tehmatt:
Fixes #297.  Seems like the fields here were accidentally copy-pasted into two different structs.


LLukas22:
Good catch!

PR #303, opened 2023-06-09
Add reference to Pythong bindings to README.md?

laurids-reichardt:
Thanks to all maintainers and contributors for their work on this great library!

Just discovered this library, but almost ended up not using it, because I assumed no Python bindings exist. That's when I discovered @LLukas22 [repo](https://github.com/LLukas22/llm-rs-python). Maybe a short reference to the Python bindings would be helpful to newcomers?

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,10 @@ Currently, the following models are supported:
   [Wizard](https://github.com/nlpxucan/WizardLM))
 - [MPT](https://www.mosaicml.com/blog/mpt-7b)
 
+Unofficial third-party language bindings:
+
+- [Python](https://github.com/LLukas22/llm-rs-python)
+
 ## Using `llm` in a Rust Project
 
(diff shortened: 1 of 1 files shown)


danforbes:
Are there Python bindings available for the reference implementation of GGML?

LLukas22:
There is [ctransformers](https://github.com/marella/ctransformers) which wraps the ggml examples directly and [llama-cpp-python](https://github.com/abetlen/llama-cpp-python) which wraps llama.cpp.

danforbes:
Can we do it like this instead (sorry I'm pedantic) https://github.com/danforbes/llm/blob/docs/readme/bindings/README.md#using-llm-from-other-languages

LLukas22:
Looks good. We should also add [llama-node](https://github.com/Atome-FE/llama-node).

danforbes:
Thank you for your contribution! I've maintained your commit on #309 💯

Issue #299, opened 2023-06-06
Support ggml metal backend


pixelspark:
Support for Metal GPU acceleration on macOS (and I assume iOS) just merged in llama.cpp master: https://github.com/ggerganov/llama.cpp/pull/1642

It would be great if this could also be employed from `llm`. I assume all it needs is something similar to https://github.com/rustformers/llm/pull/282/files and perhaps setting a flag at runtime (the `-ngl 1` parameter should be set on llama.cpp's `./main` to enable Metal).

darxkies:
I could update the PR for metal support. Could you test it together with the other backends (cuBLAS / CLBlast)?

pixelspark:
Definitely! Currently traveling but will have access to my M1 Max machine tomorrow again.

darxkies:
Unfortunately, ggerganov/ggml does not have support for Metal yet. Thus it can not be enabled in rustformers/llm yet. Never the less it would be great if you could test the RP's cuBLAS / CLBlast functionality.

pixelspark:
Hm, the Metal bits should be upstreamed to `ggerganov/ggml` soon, right? (As far as I understand it, it is wholly contained in two files).

I quickly tested CLBlast on my work laptop with llama.cpp itself, but its AMD Radeon GPU appears a bit too shitty for it to work. I will attempt a test with `cuBLAS` later on a more beefy machine with NVIDIA hardware.

darxkies:
I've updated the BLAS PR to support Metal too. You can use it like so:

```shell
cargo run --release --features metal mpt infer --model-path mpt-7b-chat-q5_1.bin -p "Once upon a time"
```

Could you please test it out?

pixelspark:
Nice 👍🏻  I am seeing the following errors at build time:

````
error: environment variable `CUDA_PATH` not defined at compile time
   --> crates/ggml/sys/build.rs:126:39
    |
126 |         let targets_include = concat!(env!("CUDA_PATH"), r"\include");
    |                                       ^^^^^^^^^^^^^^^^^
    |
    = help: use `std::env::var("CUDA_PATH")` to read the variable at run time
    = note: this error originates in the macro `env` (in Nightly builds, run with -Z macro-backtrace for more info)

error: environment variable `CUDA_PATH` not defined at compile time
   --> crates/ggml/sys/build.rs:127:35
    |
127 |         let targets_lib = concat!(env!("CUDA_PATH"), r"\lib\x64");
[...]

PR #311, opened 2023-06-13
Implement Metal support

pixelspark:
This is an attempt at adding support for Metal (#299). It is based on #282 which builds the GGML metal support (as well as CLBlast/cuBLAS). Many thanks to @darxkies for getting this to work because I have no clue about how to get these Rust build scripts to work!

Currently I have implemented 

- [X] A config flag and command line arguments for enabling the GPU (I would prefer it to be a runtime option as it is not clear whether Metal is actually much faster on macOS for now, and also that would allow using CPU and GPU in parallel and/or for different parts of the model). 
- [X] Basic support for initializing a ggml_metal_context and calls to graph compute for the `mpt` model. 
 
What still needs to be done:

- [x] Send actual tensor data to Metal ([this part in llama.cpp](https://github.com/ggerganov/llama.cpp/blob/92549202659fc23ba9fec5e688227d0da9b06b40/llama.cpp#L2617)). This step is the most difficult for me as I am not too familiar with the tensor/buffer management in `llm`. Perhaps @philpax / @LLukas22 / @darxkies can lend a hand here, it might be very easy for you actually.
- [x] Do not create and destroy a metal context all the time (creating/destroying a `Context` is apparently done quite often and `MetalContext` is currently a field of `Context`). The `MetalContext` should probably be persisted (or at least there should be a way to prevent it from having to compile the shader code each time).
- [x] Implement Metal for all the models (I will be testing with `Mpt` first)
- [x] Figure out a way to feed `ggml-metal.metal` source (ggml-metal.m currently looks for the file [using NSBundle](https://github.com/ggerganov/llama.cpp/blob/92549202659fc23ba9fec5e688227d0da9b06b40/ggml-metal.m#L122) so it needs to be in the same directory as the binary, in the future they apparently want to [embed the shader code](https://github.com/ggerganov/llama.cpp/blob/92549202659fc23ba9fec5e688227d0da9b06b40/ggml-metal.m#L78).


LLukas22:
I'm willing to help with this, but i don't own a mac and can't even compile with metal support 😅. 

>Send actual tensor data to Metal ([this part in llama.cpp](https://github.com/ggerganov/llama.cpp/blob/92549202659fc23ba9fec5e688227d0da9b06b40/llama.cpp#L2617)). This step is the most difficult for me as I am not too familiar with the tensor/buffer management in llm. Perhaps @philpax / @LLukas22 / @darxkies can lend a hand here, it might be very easy for you actually.

I can take a look at this but i can't test it. Maybe i just hack something in and push it to this PR. If it doesn't work we can just throw it out again.

pixelspark:
> I can take a look at this but i can't test it. Maybe i just hack something in and push it to this PR. If it doesn't work we can just throw it out again.

Sure! I have added you and @darxkies as collaborators to my fork so you should be able to push commits to it.

LLukas22:
I added a very hacky implementation to the `MetalContext` just to check if the right buffers are send to metal, this will probably crash after the first inference call, if it works at all. But it should show you which buffers/tensors have to be moved to metal.

pixelspark:
> I added a very hacky implementation to the `MetalContext` just to check if the right buffers are send to metal, this will probably crash after the first inference call, if it works at all. But it should show you which buffers/tensors have to be moved to metal.

@LLukas22 I am getting:

````
ggml_metal_add_buffer: allocated 'data            ' buffer, size =     0.06 MB
ggml_metal_add_buffer: allocated 'eval            ' buffer, size =  1024.00 MB
ggml_metal_add_buffer: failed to allocate 'k               ' buffer, size =     0.02 MB
Message:  Could not add k buffer to metal context
Location: crates/ggml/src/metal.rs:67

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                                ⋮ 8 frames hidden ⋮                               
   9: ggml::metal::MetalContext::initialize_buffers::hc52db6f22f029135
      at <unknown source file>:<unknown line>
  10: <llm_mpt::Mpt as llm_base::model::KnownModel>::evaluate::h6fa9c3d15352646a
      at <unknown source file>:<unknown line>
  11: llm_base::inference_session::InferenceSession::infer::h6b81faf80455e8aa
      at <unknown source file>:<unknown line>
  12: llm::main::hda091713b6405734
      at <unknown source file>:<unknown line>
````

Digging a bit further it seems that the `ggml_metal_add_buffer` function calls into `[MTLDevice newBufferWithBytesNoCopy:length:options:deallocator:]` and that expects `data` to be [page aligned](https://developer.apple.com/documentation/metal/mtldevice/1433382-newbufferwithbytesnocopy). The buffer however doesn't appear to be page-aligned (my page size on M1 is 16384 and the pointer address modulo that is not zero).

Further digging reveals: llama.cpp in `kv_cache_init` calls `cache.buf.resize(2u*n_elements*ggml_type_size(wtype) + 2u*MB);`. The implementation for this resize method aligns the buffer to the page boundary when using Metal:

````cpp
void resize(size_t len) {
#ifdef GGML_USE_METAL
        free(addr);
        int result = posix_memalign((void **) &addr, getpagesize(), len);
        if (result == 0) {
            memset(addr, 0, len);
        }
        else {
            addr = NULL;
        }
#else
        delete[] addr;
        addr = new uint8_t[len];
#endif
        size = len;
    }
````

So... how do you think this should be handled in llm?

**Comment by LLukas22** [comment]
Hm, for the time being we could just hack it in while creating the k/v-Tensors. Also the `data` offload seams a bit suspicious with it's 0.6 MB. Could you try to switch mmap off? 

Basically i would like to get metal working even if it's hacky just to see if we need bigger changes somewhere. I'm currently working on the CuBLAS implementation and already notices some stuff that needs changing. If we have both implementations we can pull them together and clean up the mess.

**Comment by pixelspark** [comment]
> Hm, for the time being we could just hack it in while creating the k/v-Tensors. Also the `data` offload seams a bit suspicious with it's 0.6 MB. Could you try to switch mmap off?

With `--no-mmap` the size for `data` looks better, but the error remains:

````
newBufferWithBytesNoCopy: 10737418240 aligned_size=-553762816
ggml_metal_add_buffer: allocated 'data            ' buffer, size =  3567.89 MB
newBufferWithBytesNoCopy: 16089333760 aligned_size=1073741824
ggml_metal_add_buffer: allocated 'eval            ' buffer, size =  1024.00 MB
note, data buffer is not page-aligned? page_size=16384newBufferWithBytesNoCopy: 14478622976 aligned_size=16384
ggml_metal_add_buffer: failed to allocate 'k               ' buffer, size =     0.02 MB
Message:  Could not add k buffer to metal context
Location: crates/ggml/src/metal.rs:67
````

**Comment by LLukas22** [comment]
Yeah the data offload now looks correct, regarding the k/v error the tensors are created [here](https://github.com/rustformers/llm/blob/main/crates/llm-base/src/inference_session.rs#L440-L441). I don't know how to align memory in rust so maybe just hack something together? 😅

**Comment by pixelspark** [comment]
> Yeah the data offload now looks correct, regarding the k/v error the tensors are created [here](https://github.com/rustformers/llm/blob/main/crates/llm-base/src/inference_session.rs#L440-L441). I don't know how to align memory in rust so maybe just hack something together? 😅

See my comment above on how to allocate aligned memory! (Now all I need is a proper way to find out the page size). Your link explains why my fix didn’t work - seems there are actually two places that allocate `memory_k/v` and I missed the clone impl.

**Comment by louisgv** [comment]
> Now all I need is a proper way to find out the page size

I went thru some digging on this and found a lot of conflicting answer in the unix world. There's a solution that use a sysconf: https://github.com/Elzair/page_size_rs/blob/master/src/lib.rs#L100

But further reading on what this actually does shown it might return a static number instead of the actual available page_size: https://stackoverflow.com/questions/21159982/how-to-get-the-page-size-of-a-specific-address-programmatically

I'm not sure if that'd suffice tho? Do we want the optimal page size allocated at the time, or do we want the maxium page size that can be allocated?

**Comment by pixelspark** [comment]
> > Now all I need is a proper way to find out the page size
> 
> I went thru some digging on this and found a lot of conflicting answer in the unix world. There's a solution that use a sysconf: https://github.com/Elzair/page_size_rs/blob/master/src/lib.rs#L100
> 
> But further reading on what this actually does shown it might return a static number instead of the actual available page_size: https://stackoverflow.com/questions/21159982/how-to-get-the-page-size-of-a-specific-address-programmatically
> 
> I'm not sure if that'd suffice tho? Do we want the optimal page size allocated at the time, or do we want the maxium page size that can be allocated?

Yeah well all that is really needed is the page size on macOS (or specifically the alignment requirement for Metal buffers) so anything that works on (just/at least) macOS will do. For now we could even hard code 16384 for ARM (and 4096 for x64) as this is unlikely to change soon.

**Comment by LLukas22** [comment]
@pixelspark Any drawbacks on always aligning with the page size on non mac-os platforms? If not we could just always align the tensors and call it a day.

**Comment by louisgv** [comment]
> For now we could even hard code 16384 for ARM (and 4096 for x64) as this is unlikely to change soon.

This seems fine to me IMO.

**Comment by LLukas22** [comment]
@pixelspark Correct me if i'm wrong, but basically we need a function that we can point to an exact memory region and it just creates an 1d tensor out of this memory region?

**Comment by LLukas22** [comment]
@pixelspark Wow good job, regarding the error on second inference call. Thats expected as is just put the buffer init function there to show which tensors need to be assigned to metal. We should split this up. The `data` context should be offloaded in the new function of the model. The k/v and scratch tensors maybe during the new function of the session. And lastly the `ctx0` has to be offloaded every inference call. (maybe we should change that to only be offloaded once). 

But the good thing is we already passed a forward call, which means this should probably work if we fix it to assign the buffers only once :D

**Comment by pixelspark** [comment]
> @pixelspark Wow good job, regarding the error on second inference call. Thats expected as is just put the buffer init function there to show which tensors need to be assigned to metal. We should split this up. The `data` context should be offloaded in the new function of the model. The k/v and scratch tensors maybe during the new function of the session. And lastly the `ctx0` has to be offloaded every inference call. (maybe we should change that to only be offloaded once).
> 
> But the good thing is we already passed a forward call, which means this should probably work if we fix it to assign the buffers only once :D

OK so I just pushed my cleaned-up changes. The next thing to do would be to move `MetalContext` out of `Context` and put it in `InferenceSession` or some other place (it could be shared between sessions even?). Could you also have a go at doing what you write above on top of my changes?

**Comment by LLukas22** [comment]
>OK so I just pushed my cleaned-up changes. The next thing to do would be to move MetalContext out of Context and put it in InferenceSession or some other place (it could be shared between sessions even?). Could you also have a go at doing what you write above on top of my changes?

If i'll find some time i give it a try, but currently i'm focused on getting cublas working for all architectures. And debugging metal is pretty inconvenient for me as i can't build or test it. But we already know it should theoretically work, which is nice. Maybe @louisgv  could help out a bit here.

**Comment by louisgv** [comment]
@LLukas22 @pixelspark I can try to hack out this part:

> And lastly the ctx0 has to be offloaded every inference call. (maybe we should change that to only be offloaded once).

Tho, @pixelspark how are you running the test and debug perf ATM? I need some help setting those up. RN I just build your branch with `--flag metal` just to confirm it build, then run it off the cli. Was wondering if you have some command under your history that you could share so I can run and debug the implementation

**Comment by pixelspark** [comment]
> @LLukas22 @pixelspark I can try to hack out this part:
> 
> > And lastly the ctx0 has to be offloaded every inference call. (maybe we should change that to only be offloaded once).
> 
> Tho, @pixelspark how are you running the test and debug perf ATM? I need some help setting those up. RN I just build your branch with `--flag metal` just to confirm it build, then run it off the cli. Was wondering if you have some command under your history that you could share so I can run and debug the implementation

This is my oneliner:

````sh
RUST_BACKTRACE=1 cargo run --release --features metal mpt infer --use-gpu --model-path ~/Downloads/models/mpt-7b-chat-q4_0-ggjt.bin -p "Once upon a time" --no-mmap
````

Models [here](https://huggingface.co/rustformers/mpt-7b-ggml). 

````sh
shasum ~/Downloads/models/mpt-7b-chat-q4_0-ggjt.bin 
fa69fe40240cf6b0e951171f653e75881f332600 
````

**Comment by louisgv** [comment]
@pixelspark How do you load the `ggml-metal.metal` file? I tried to copy it manually to where I have the model, as well as the location of running the cargo run, but both gave me this err:

```
     Running `target/release/llm mpt infer --use-gpu --model-path '/Users/lab/Documents/LLM MODELS/MPT 7B Chat.bin' -p 'Once upon a time' --no-mmap`
⣾ Loaded hyperparametersggml_metal_init: allocating
ggml_metal_init: using MPS
ggml_metal_init: loading '(null)'
ggml_metal_init: error: Error Domain=NSCocoaErrorDomain Code=258 "The file name is invalid."
```

pixelspark:
I just copied it to the `target/release` folder right next to the llm binary. If that fails then you say want to uncomment  the printf statements near the NSBundle call in ggml-metal.m to see where it is expecting the file.

pixelspark:
OK, baby steps, but we're getting there 🐌 

I just refactored the code a bit so we use one `MetalContext` per call to `evaluate`. In this context I load all the relevant buffers (some of which are persisted in `InferenceSession`). This seemed to work great (I also added some logic to prevent deallocation of these contexts while buffers from it are still in use by the `MetalContext`).  I then run into the issue that `ggml_metal_get_buffer` was unable to find buffers that are from another context... So I moved the scratch buffers into `ctx0`. Another issue I encountered is that `NORM` op is not implemented on Metal, so we need to test with a LLaMA model instead of an MPT model for now. The incantation to use:

````bash
RUST_BACKTRACE=1 cargo run --release --features metal llama infer --use-gpu --model-path ~/Downloads/models/ggml-vicuna-7b-1.1-q4_0.bin -p "Once upon a time" --no-mmap
````

In the current state it seems to be calculating the full model in Metal! 

image

However, I still get an error:

````
[...]
gml_metal_get_buffer: error: buffer is nil
ggml_metal_graph_compute: op - MUL_MAT
ggml_metal_graph_compute: src0 - q4_0 [ 4096, 32000,     1], 1, leaf_568
ggml_metal_graph_compute: src1 -  f32 [ 4096,     1,     1], 1, node_1186
ggml_metal_graph_compute: dst  -  f32 [32000,     1,     1], 1, node_1187
ggml_metal_graph_compute: time elapsed = 1758.961125 ms
ggml_metal_get_tensor: extract results for tensor 'node_1187'
ggml_metal_free: deallocating
,The application panicked (crashed).
Message:  WeightedIndex error: InvalidWeight
Location: crates/llm-base/src/samplers.rs:157

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                                ⋮ 9 frames hidden ⋮                               
  10: <llm_base::samplers::TopPTopK as llm_base::samplers::Sampler>::sample::h4c19b36bf0aa3628
      at <unknown source file>:<unknown line>
  11: llm_base::inference_session::InferenceSession::infer_next_token::ha9360b9a5a32a6ea
      at <unknown source file>:<unknown line>
  12: llm_base::inference_session::InferenceSession::infer::hfbbb33a98974a10e
      at <unknown source file>:<unknown line>
  13: llm::main::hda091713b6405734
      at <unknown source file>:<unknown line>
  14: std::sys_common::backtrace::__rust_begin_short_backtrace::h53e9ebf55df25e67
      at <unknown source file>:<unknown line>
  15: std::rt::lang_start::{{closure}}::h0c97942b54546e14
      at <unknown source file>:<unknown line>
  16: core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &F>::call_once::h6f7eb9f266759f90
      at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ops/function.rs:287
  17: std::panicking::try::do_call::h54b2febb9ea02379
      at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/panicking.rs:485
  18: std::panicking::try::h95a2f9f45aeb75ea
      at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/panicking.rs:449
  19: std::panic::catch_unwind::h9686256fa0fc97a1
      at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/panic.rs:140
  20: std::rt::lang_start_internal::{{closure}}::h227e8b10bc4e486b
      at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/rt.rs:148
  21: std::panicking::try::do_call::h414d500a3ee5fa44
      at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/panicking.rs:485
  22: std::panicking::try::h4f025820961f1c3f
      at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/panicking.rs:449
  23: std::panic::catch_unwind::h0b71dfe3538d125d
      at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/panic.rs:140
  24: std::rt::lang_start_internal::h8ee16b8f6c950a26
      at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/rt.rs:148
  25: _main<unknown>
      at <unknown source file>:<unknown line>

Run with COLORBT_SHOW_HIDDEN=1 environment variable to disable frame filtering.
Run with RUST_BACKTRACE=full to include source snippets.
````

As you can see it sometimes complains that it cannot find a buffer (this can be fixed by moving the relevant tensor to the `ctx0` context). This is likely the reason for the fault at the end. The fix hence is to add the `data` buffer to the evaluation context (but I'm not yet sure how). 

Full log https://gist.github.com/pixelspark/8beffe4121bbed69f1ac272ef26a4f5b#file-gistfile1-txt

(Note, even though `GGML_METAL_NDEBUG` is defined for some reason I did not get debugging logs until editing `ggml-metal.m` by hand to enable it).

pixelspark:
So the weight tensors themselves are in `Llama::layers` which are allocated in `Llama::context`. Just having one big context and adding that context's buffer to the `MetalContext` would likely make this work as `ggml_metal_get_buffer` can index into Metal buffers (the only requirement would be for the memory of `LLama::context` to be 16k-aligned). 

However, we use split contexts (one for the model and at least one for each inference session). Therefore we really need a change in `ggml_metal_get_buffer` so that it allows looking into other contexts. Or do you guys see alternative solutions?

pixelspark:
OK so as of the most recent commit it also loads the model weights into the Metal context properly. It is still missing some buffers, will look into this later. I have a feeling we're close :-)

pixelspark:
After rebasing to the most recent version and making some final changes, this now seems to somewhat work. Issues:

- Output is garbage after the first token. Something may be up with persisting memory_k/v between runs?
- It does something weird when mmap is enabled.
- There are still a zillion `ggml_metal_get_buffer: error: buffer is nil` warnings. I figured this may not be an issue as they appear to be related to intermediate outputs of operations like 'view' and 'copy' which are no-ops in Metal. Then again I am not seeing these errors with `llama.cpp` itself so something might still be wrong.
- We are creating a new Metal context for each call to `evaluate`. The `EvaluationContext` should be saved between calls to `evaluate`. This however is difficult as the current code creates a new context on each call to `evaluate` (in particular it allocates an `embd` tensor whose size depends on the number of input tokens). 
- Due to the above the following warning appears many times (it appears ggml-metal does not really deallocate Metal resources): `warning: current allocated size is greater than the recommended max working set size`

I don't have any further time today but feel free to hack on this a bit more.

pixelspark:
Did manage to find some time (this is starting to look like a fun monologue :-)). Long story short, it now works!

image


Final thing to do before this can be merged is to fix mmap and to put the `EvaluationContext` in the `InferenceSession` so we don't recreate it each time.

LLukas22:
Wow good job! Didn't expect this to be implemented that quickly o_O

What caused the output to be garbled? I think i'm facing a similar issue with the CuBLAS implementation. 
Furthermore i also noticed that some GGML operations are not implemented in the Metal/CuBLAS/CLBLAST kernels. Currently im building something to put certain unsupported operations automatically back on the CPU. Could we exclude specific operations from being assigned to the metal buffer and execute them on the CPU? Optimally i would like to make every model "just work" on any accelerator but i think first focusing on LLama is a good first step.

Regarding mmap, i would just disable it if a metal accelerator is used, as we have to read the tensors either way into RAM. But i think @philpax knows more about that than me. 

I also need the `EvaluationContext` to be part of the `InferenceSession` for the CuBLAS implementation as i only want to offload the evaluation buffer once. I'll take a look at what you changed and will compare it against my CuBLAS/CLBLAST implementations.
Hopefully we can implement them a seamlessly as possible.

pixelspark:
> Wow good job! Didn't expect this to be implemented that quickly o_O

Me neither ;-)

> What caused the output to be garbled? I think i'm facing a similar issue with the CuBLAS implementation. 

I allocated my own aligned buffers for memory_k/v (see many comments back). However, because this memory fell outside of the memory allocated to the context of these tensors, GGML couldn't map them to Metal buffers. Turns out that the aligned allocation was not necessary after all (you can just add the whole `ggml_context` as Metal buffer and it will read individual tensors from that). Much easier.

> Furthermore i also noticed that some GGML operations are not implemented in the Metal/CuBLAS/CLBLAST kernels. Currently im building something to put certain unsupported operations automatically back on the CPU. Could we exclude specific operations from being assigned to the metal buffer and execute them on the CPU? Optimally i would like to make every model "just work" on any accelerator but i think first focusing on LLama is a good first step.

Not sure how to implement that...

> Regarding mmap, i would just disable it if a metal accelerator is used, as we have to read the tensors either way into RAM. But i think @philpax knows more about that than me.

Just give me five minutes 😄 

> I also need the `EvaluationContext` to be part of the `InferenceSession` for the CuBLAS implementation as i only want to offload the evaluation buffer once. I'll take a look at what you changed and will compare it against my CuBLAS/CLBLAST implementations. Hopefully we can implement them a seamlessly as possible.

:+1:

pixelspark:
So mmap works too now. In 55a8e5532928826d8e365c8b45512dc32541babe I moved the `Mmap` struct to `Context` which makes the handling much cleaner (models don't need to keep hold of the `Mmap` struct anymore).

LLukas22:
Sounds good!

> In [55a8e55](https://github.com/rustformers/llm/commit/55a8e5532928826d8e365c8b45512dc32541babe) I moved the `Mmap` struct to `Context`, which makes the handling much cleaner (models don't need to keep hold of the `Mmap` struct anymore).

Nice, getting mmap to work and cleaning up how we handle the `Mmap` struct is a big win! Plus, it sounds like we're now more in line with the behavior of the CuBLAS/CLBLAST implementations. I'll take a closer look at the changes when I have time tomorrow.

> Not sure how to implement that...

You're the wgpu whiz here, I thought you might have some ideas! 😄 But don't worry, I'm finding the CuBLAS implementation a bit tricky too. We can give it another go when we've got more time.

So, it looks like the last piece of the puzzle is to include the `ggml-metal.metal` file, right?

pixelspark:
Well the last remaining puzzle piece is being able to re-use a built graph (currently each call to `evaluate` will build a new graph in `ctx0`, but we can't do that with Metal since resources can only be deallocated by deallocating the entire Metal context, but we want to keep that alive).

I have something in the works (it also cleans up the model interface a bit) but it requires a few more hours, will be later this week hopefully..

pixelspark:
Finally: Metal support is ready! 🎉  Below is Vicuna-13B on an M1 Max (the GIF makes it look like it generates a few tokens at a time, in reality it is just one token at once, but of course at the same speed):

metal_llm

The cbbe41c25ba222e42fa3e542d457e992006b0735 commit refactors the way `InferenceSession` and the various models interact while building a model graph (it retains the underlying buffers in memory). The interface is greatly simplified but still relatively safe. Models need not even be aware of whether or not `mmap` or the GPU is enabled!

I was successfully able to run LLaMA models on CPU and GPU, both with and without mmap enabled. I did encounter some issues chatting with `mpt` (it works but leads to a `WeightError` after a while). This might be due to issues in `llama.cpp` itself. As GPU support can simply be disabled for these models (i.e. by not passing `--use-gpu`) I think we should merge this anyway (also so that @LLukas22 can use this for his cuBLAS/CLBlast work). 

The API changes from the outside are minimal - users will have to pass a new `use_gpu` flag and `InferenceSession` does not implement `Clone` anymore (it is possible to implement it but the semantics should be made clear - does it create a new fresh session reusing resources, does it clone but including the memory, etc.). 

Some remarks for the Metal version:

philpax:
Fantastic work! I'll take a look soon 🚀

louisgv:
@pixelspark amazing works - sorry I haven't been able to contribute since, got heavily distracted by errands offline

PR #324, opened 2023-06-21
fix(metal): Copy ggml-metal.metal to target directory

nightscape:
Rust noob here :wink:
I've hacked a way to copy the required `ggml-metal.metal` file into the `target/<profile>` directory, otherwise users would always have to do it by themselves.
Not sure if there's a nicer way, especially the `Go up 3 directories` part rubs me the wrong way.
Still better than having a semi-broken build imo.

### DIFF
--- a/crates/ggml/sys/build.rs
+++ b/crates/ggml/sys/build.rs
@@ -1,5 +1,5 @@
 use std::env;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 // By default, this crate will attempt to compile ggml with the features of your host system if
 // the host and target are the same. If they are not, it will turn off auto-feature-detection,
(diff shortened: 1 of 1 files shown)


LLukas22:
@pixelspark What are your thoughts about this? I have absolutely no idea about the metal stuff, so i won't comment on this 😅

pixelspark:
The current situation is temporary - at some point the shader code will be embedded in `ggml-metal.c` (using the equivalent of Rust's `include_str!`) - at least that is the intention indicated in the code. 

For now I think this PR adds some value. Please add a comment/message somewhere as a reminder that this can be removed when `ggml-metal.metal` is not found anymore (and/or add an issue to keep track of this).

nightscape:
Added a comment in the code linking to the discussion here.
@pixelspark do you have a link for the corresponding discussion so we can reference it?

Good to merge from my side 😃

pixelspark:
See here: https://github.com/ggerganov/llama.cpp/blob/7487137227eb32ed9b12156338b865cb29b2dfd1/ggml-metal.m#L80

philpax:
Thanks for the PR! Yeah, this is better than having a build that doesn't work as expected.

However, I suspect this may not work as well if the build setup is non-standard, and the resulting file's likely to be dropped as part of deployment. It doesn't matter too much because this should be fixed upstream by being included in the actual file, but still...

Let me see if I can think of an alternate fix that might be a little more robust.

PR #325, opened 2023-06-22
CUDA/OpenCL Acceleration

LLukas22:
Implements CUDA/OpenCL acceleration via CuBLAS/CLBLAST. 

https://github.com/rustformers/llm/assets/65088241/4733a6c3-2b85-4d0b-9428-aa4160f41d5f

Stuff that works:
- [x] Offload the weights to the GPU
- [x] Enable gpu acceleration via `--use-gpu`
- [x] Control how many layers are offloaded via `--gpu-layers`

Stuff that still needs to be done:

- [x] Check why CLBLAST's matmul throws an error in the current llama.cpp version.
- [x] Decide on how we want to handle other architectures
- [x] Find another way to pass a mutable `ctx0` (I dont want to use a RefCell here)
- [ ] Find  a way to pass additional arguments to ggml-sys's `build.rs` to enable `f16` [optimizations](https://github.com/ggerganov/llama.cpp/commit/16b9cd193965769089881bb8ec012fccca7b37b6). Maybe @darxkies could help here. 
- [x] Find a better place to initialize the gpu context and cuda scratch buffer

Nice to have:
- [ ] Implement Multi-GPU support
- [x] Benchmark against llama.cpp (Maybe something for @jafioti ?)


jafioti:
Awesome to see the progress! I'll run some benchmarks on my machine (2080 super) tonight

LLukas22:
>     * [x]  Decide on how we want to handle other architectures

We will first focus on supporting only LLama. Other architectures will be supported via commites to the llama.cpp/ggml repo. Which implement the missing ggml_ops in CUDA/OpenCL.

jafioti:
Some quick benchmarks:
GPU: 2080 Super
Model: Llama 7B
Prompt: "Rust is a cool programming language because"
Num Tokens Generated: 128
llm Command: `cargo run --features cublas --release llama infer -m ../llama.cpp/models/llama-7b.ggmlv3.q4_0.bin -p "Rust is a cool programming language because" --stats --use-gpu --gpu-layers 100 --num-predict 128`

This Branch Stats:
```
feed_prompt_duration: 317ms
prompt_tokens: 9
predict_duration: 9548ms
predict_tokens: 137
per_token_duration: 69.693ms
```
llama.cpp (commit d7b7484f74d486f77feb4c0b7af7e1718ed91651) Stats:
```
llama_print_timings:        load time =  1015.55 ms
llama_print_timings:      sample time =    52.68 ms /   128 runs   (    0.41 ms per token)
llama_print_timings: prompt eval time =   182.96 ms /     9 tokens (   20.33 ms per token)
llama_print_timings:        eval time =  2901.68 ms /   127 runs   (   22.85 ms per token)
llama_print_timings:       total time =  3162.59 ms
```

LLukas22:
@jafioti I can't reproduce your results i did my own benchmarking with the following setup and results.

Device:
Windows 11 
RTX 3090

Models:
- [OpenLLama 3B Q4_0](https://huggingface.co/rustformers/open-llama-ggml/resolve/main/open_llama_3b-q4_0-ggjt.bin)
- [OpenLLama 7B Q4_0](https://huggingface.co/rustformers/open-llama-ggml/resolve/main/open_llama_7b-q4_0-ggjt.bin)
- [Nous Hermes 13b Q5_1](https://huggingface.co/TheBloke/Nous-Hermes-13B-GGML/resolve/main/nous-hermes-13b.ggmlv3.q5_1.bin)
- [Wizard-Vicuna-Uncensored 30B Q5_1](https://huggingface.co/TheBloke/Wizard-Vicuna-30B-Uncensored-GGML/resolve/main/Wizard-Vicuna-30B-Uncensored.ggmlv3.q5_1.bin)


commands:

llm: `cargo run --features cublas --release llama infer -m [MODELPATH] -p "Rust is a cool programming language because" --stats --use-gpu --gpu-layers 100 --num-predict 128`

llama.cpp: `.\main.exe -m [MODELPATH] -p "Rust is a cool programming language because" -n 128 -ngl 100`

(For Wizard-Vicuna-Uncensored 30B only the first 40 layers were offloaded to make the model fit into 24 GB Vram and the token limit was reduced from 128 to 50 tokens)

Results:
6 runs were performed for each model, the results of the first run were discarded and the other 5 runs were averaged.

The following table shows the per token duration:

|                    	| llama.cpp 	| llm 	|
|--------------------	|-----------	|-----	|
| OpenLLama 3B       	|      N/A     	|    16.17 ms 	|
| OpenLLama 7B       	|         17.54 ms  	|   19.40 ms 	|
| Nous Hermes 13B    	|          41.23 ms 	|    42.72ms 	|
| Wizard Vincune 30B 	|         312.30 ms  	|    289.97ms 	|


`llm` seams to be about 1-2 ms slower than `llama.cpp` which is probably caused by different measuring locations, i think `llm` includes the token callback into the measurement.

jafioti:
I've done another test on an A10 with [OpenLLama 7B Q4_0](https://huggingface.co/rustformers/open-llama-ggml/resolve/main/open_llama_7b-q4_0-ggjt.bin)

I did the same, 6 tests, keep last 5
Results:

llama.cpp | llm
-- | --
OpenLLama 7B | 21.44 ms | 28.02 ms

It should be noted that I did see a wider variance in the llm numbers. Here are the raw observed numbers:
llm: `30.8, 22.3, 32.5, 32, 22.5`
llama.cpp: `21.5, 21.5, 21.3, 21.4, 21.5`

LLukas22:
OK thats very interesting, any idea on why the performance of the 2080 was that much worse? 


The A10 results seam to be ok'ish. I still have to check how we meassure token times and how LLama.cpp does it.

jafioti:
No idea, will check on an a100 and see if the gap shrinks furthur

LLukas22:
Could it be possible for the 2080 to run out of memory and start paging into RAM? `llm` seams to allocate a bit more VRAM than `llama.cpp` 🤔

jafioti:
Alright another run on an A100 using [Wizard 33B](https://huggingface.co/TheBloke/WizardLM-33B-V1.0-Uncensored-GGML):



Model | llama.cpp | llm
-- | -- | --
WizardLM 33B | 50.98 ms | 83.79 ms

Interestingly this discrepency is mostly due to a few tokens generated on a single run in llm. Here are the runtimes of each llm run: `52.47, 56.07, 192.3, 61.2, 57.4`. As you can see, one of them took a lot longer, and when running I noticed most of the tokens generated fast, but like 5 of them took almost a whole second to generate. Does it decide to use CPU for some forward passes? How can I force it to always use GPU?

LLukas22:
Ok that indeed is interesting, thanks for helping by testing different cards. Theoretically all layers should be offloaded except for the embedding layer. You could omit the `--gpu-layers 100` command as `llm` will offload all layers if you don't define it and gpu acceleration is enabled. I actually don't know why the inference would slowdown for some tokens, maybe it's a problem with the sampler/tokenizer? I'll try to run some benchmarks with the huggingface tokenizer enabled to see if that changes something. If you want to test it you can define an external tokenizer via the `-r [HF/repo]` parameter. E.g. `-r "ehartford/WizardLM-33B-V1.0-Uncensored"`.

>Also for the 2080 I think if it ran out of VRAM it would just error, no?

Depending on your driver version, some drivers decide to offload into RAM if you are a bit over your available VRAM. But i only encountered it once and i don't know if that's just a windows thing. 🤔

jafioti:
Alright, really strange but I got very similar generation performance the higher I go in card power. On an A10 or H100, the token generation time is nearly identical, so it might just be my 2080, idk.

But another thing I noticed that's pretty major is the prompt feeding stage taking quite a bit longer than llama.cpp. On your branch, is the initial prompt feeding happening on GPU? Or is only the subsequent token generation offloaded?

Here's the prompt I ran through WizardLM-30B:
```
A chat between a curious user and an artificial intelligence assistant.
The assistant gives accurate answers to the user's questions while being short and to-the-point.
DOCUMENT START

Company Dividends

The table below provides a breakdown of dividends received by shareholders for Acme, Inc. and Newtech Corp. over a period of five years.

Dividends per Share
\```csv
Year,Acme, Inc.,Newtech Corp.
2016,$1.10,$0.45
2017,$1.25,$0.60
2018,$1.40,$0.75
2019,$1.55,$0.90
2020,$1.75,$1.05
\```

DOCUMENT END

USER: Please answer the following question based on the document provided above. In 2019, how much was Acme, Inc.'s dividend per share?
ASSISTANT:
```

And the prompt-only times:
llama.cpp: `580.72 ms /   231 tokens (    2.51 ms per token,   397.78 tokens per second)`
llm: `9619ms/ 231 tokens( 41.64ms per token, 24.01 tokens per second)`

LLukas22:
@jafioti Thanks for the additional tests. The prompt feeding happens on the GPU, as it's the same forward call as the inference of new tokens. The difference in feeding times is probably caused by the default `batch-size`. I think llama.cpp uses 256 or 512 as a default when running gpu inference. We curretnly always default to 8. This means  your 231 token long prompt needs only one forward call in llama.cpp but 231 // 8 forward calls on this branch. You could try increasing it via the `--batch-size` parameter.


If i get home from work im gonna sync this branch with the newest ggml source from llama.cpp and im gonna do some benchmarking to test the impact of the `batch-size` parameter.

jafioti:
Yup you were exactly right, the batch size was the key! The same prompt now takes 672ms

LLukas22:
ClBlast currently fails on Windows, see https://github.com/ggerganov/llama.cpp/issues/2065

jafioti:
@LLukas22 Can you pull main again? PR #339 was just merged and I'd like to run in an OpenSSL-free environment

sonic182:
A run in my RTX 2060

```
llm infer  -a llama -m models/open_llama_7b-q5_1-ggjt.bin -p "42: The answer to life, the universe and everything" -t 6 --stats --use-gpu --temperature 0.2 --gpu-layers 24 --num-predict 100
⣾ Loaded hyperparametersggml_init_cublas: found 1 CUDA devices:
  Device 0: NVIDIA GeForce RTX 2060
✓ Loaded 291 tensors (5.1 GB) after 1734ms
42: The answer to life, the universe and everything
The Hitchhiker's Guide to the Galaxy is a science fiction comedy series created by Douglas Adams. It was first broadcast on BBC Radio 3 in March 1978 with the original cast of Simon Jones as Arthur Dent, Peter Jones as Ford Prefect, Geoffrey McGivern as Trillian and Mark Wing-Davey as Zaphod Beeblebrox.
The series is based around the adventures of Arthur Dent (Simon Jones) after he was pl

feed_prompt_duration: 929ms
prompt_tokens: 13
predict_duration: 11517ms
predict_tokens: 113
per_token_duration: 101.920ms
```
Just a bit of info if you need it :+1:

LLukas22:
The clblast shader bugs are now fixed and inference works.

https://github.com/rustformers/llm/assets/65088241/1bd343ce-6a9f-4f93-8bf2-d38bb5c0d62a

jafioti:
@LLukas22  I'm getting a memory leak now after multiple runs. The only thing I do is start a new inference session each time. Should this not be done, or should something else be called in between runs to free up the memory used in that run?

LLukas22:
@jafioti Good point, i forgot to free the memory after a session is disposed, sorry.🤦

LLukas22:
Theoretically the memory should now be freed if the session/model is dropped but i need to test this further.

LLukas22:
Yeah i can confirm, the non k_quants seam too be broken, probably because of this [commit](https://github.com/ggerganov/llama.cpp/commit/924dd22fd3ba93e097f8d19ba5cda919ca2fe2fb#diff-66b17223e8ba54054fb2600ecbd31107f8b917bac36c7f3789811b0f0e9802a1). I'll later try to update or peg `llama.cpp` to a version where the normal quants work.

jafioti:
I've tried every quant on https://huggingface.co/TheBloke/WizardLM-13B-V1.1-GGML/tree/main and am facing the same issue, so it's not only constrained to non-K quants.

LLukas22:
I tested some of my local models and the behavior is very weird. Orca7B_V2_7B_Q5_K_M seams to work, but Wizard7B_Q5_K_M which is based on the same base model doesn't work and i don't know why. I'll try to upgrade the `llama.cpp` version in the next few days but there were a lot of changes in the graph-planning/-compiling stages and i can't use an older `llama.cpp` version as it doesn't contain the shader fixes i need for opencl 😞

philpax:
Had a *very* cursory look and this is really impressive. I'll test this out and review it properly soon.

crates/ggml/src/context.rs
```diff
@@ -70,15 +75,31 @@ impl Context {
             ptr: Arc::new(NonNull::new(raw).expect("Should not be null")),
             mmap: None,
             buffer: None,
+            can_offload: false,
         }
     }
 
+    /// If offloading is enabled, all tensors created by this context will be offloaded to the GPU
+    pub fn enable_offloading(&mut self) {
```
philpax, line 83: Hm, if the only thing that's being changed here is the boolean, I'd suggest changing this to `set_offloading_enabled` or similar. The separate methods make me think that there's additional work being done by each function.

crates/ggml/src/lib.rs
```diff
@@ -29,6 +29,65 @@ mod tests;
 #[cfg(feature = "metal")]
 pub mod metal;
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+///Accelerators supported by `ggml`.
```
philpax, line 33: Very minor style nit: `/// ` with a space after the slashes, but not a huge issue. (I'll see if rustfmt has a way to do that automatically.)

crates/ggml/src/tensor.rs
```diff
@@ -15,6 +15,36 @@ impl Tensor {
     /// Exposed for purposes of determining context size.
     pub const C_TYPE_SIZE: usize = std::mem::size_of::<sys::ggml_tensor>();
 
+    ///Sets the name of the tensor
+    pub fn set_name(&mut self, name: &str) -> &Tensor {
+        assert!(name.len() <= 48, "Name is too long!");
```
philpax, line 20: Can you use `GGML_MAX_NAME`?

crates/ggml/src/tensor.rs
```diff
@@ -15,6 +15,36 @@ impl Tensor {
...
+        let bytes = name.as_bytes();
+        let mut array = [0i8; 48];
+        array[..bytes.len()].copy_from_slice(&bytes.iter().map(|&x| x as i8).collect::<Vec<_>>());
+
+        unsafe { self.ptr.as_mut().name = array }
+        self
+    }
+
+    ///Gets the name of the tensor
+    pub fn get_name(&self) -> String {
```
philpax, line 31: just `name`, ditto with `get_backend` - Rust house style appears to be to use function names that are the same as the variable for getters

crates/llm-base/src/inference_session.rs
```diff
@@ -26,6 +26,24 @@ fn scratch_buffers() -> ScratchBuffers {
     ]
 }
 
+fn kv_memory(
```
philpax, line 29: Helper functions at the bottom; the first thing a user should see looking at the file is the public interface.

crates/llm-base/src/loader.rs
```diff
@@ -670,6 +672,15 @@ impl TensorLoader<LoadError> for MmapCompatibleLoader<'_> {
         Ok(tensor)
     }
 
+    fn offload(&mut self, name: &str, backend: ggml::Backend) -> Result<ggml::Tensor, LoadError> {
```
philpax, line 675: Maybe a `transfer` method on the `Tensor` that can be called as required.

crates/llm-base/src/loader.rs
```diff
@@ -741,6 +752,15 @@ impl<'a> FileContext<'a> {
             }
         }
 
+        // The tensor name is truncated to 32 characters.
+
+        let tensor_name = if name.len() > 32 {
```
philpax, line 757: 32 or 48?

crates/llm-base/src/lora.rs
```diff
@@ -126,7 +126,10 @@ impl LoraAdapter {
 
         // Compute the graph
         gf.build_forward_expand(&output);
-        patch_context.graph_compute(&mut gf);
+
+        //TODO: maybe pass the model's thread count to this context
+        let mut plan = GraphExecutionPlan::new(&mut gf, 8);
```
philpax, line 131: Yes, I'd like to resolve the matter of implicit thread counts for non-inference operations. I wonder what the best approach would be there - maybe we should just add a `ExecutionParameters { thread_count, backend }` struct as a parameter to each function? 🤔

crates/llm-base/src/model/mod.rs
```diff
@@ -207,6 +204,8 @@ pub struct ModelParameters {
     pub lora_adapters: Option<Vec<PathBuf>>,
     /// Whether to use GPU acceleration when available
     pub use_gpu: bool,
+    /// The number of layers to offload to the gpu. If `None`, all layers will be offloaded.
```
philpax, line 207: I was confused at first - I thought that setting this to `None` would always offload all layers to the GPU - but then I saw the `use_gpu` above. I'd suggest altering the comment to mention that the layers will only be offloaded if `use_gpu` is set.

crates/llm/examples/drop_multiple_sessions.rs
```diff
@@ -0,0 +1,57 @@
+use llm::ModelArchitecture;
+use llm_base::{InferenceFeedback, InferenceParameters, ModelParameters};
+use std::{convert::Infallible, path::PathBuf};
+
+fn main() {
+    let prompt = "What is the meaning of life?";
+    let model_path = PathBuf::from(r"C:\Users\lkreu\Downloads\orca-mini-v2_7b.ggmlv3.q5_K_M.bin");
```
philpax, line 7: Use clap, like the other examples, to avoid hardcoding a path

crates/llm/examples/drop_multiple_sessions.rs
```diff
@@ -0,0 +1,57 @@
...
+        session
+            .feed_prompt(model.as_ref(), prompt, &mut Default::default(), |_| {
+                Ok::<InferenceFeedback, Infallible>(llm::InferenceFeedback::Continue)
+            })
+            .unwrap();
+        drop(session);
+        println!("Dropped session {i}");
+    }
+
+    drop(model);
```
philpax, line 39: I think this could probably be an integration test?

We could also potentially use something like [mockalloc](https://docs.rs/mockalloc/latest/mockalloc/) to automatically check we've freed all the memory we've allocated 🤔

crates/models/llama/src/lib.rs
```diff
@@ -34,11 +34,22 @@ pub struct Llama {
...
     context: Arc<ggml::Context>,
+    loaded_tensors: HashMap<String, ggml::Tensor>,
 }
 
 unsafe impl Send for Llama {}
 unsafe impl Sync for Llama {}
 
+impl Drop for Llama {
+    fn drop(&mut self) {
+        for (_, tensor) in self.loaded_tensors.drain() {
```
philpax, line 45: I'd suggest turning `loaded_tensors` into a struct that wraps the tensors, and does the transfer on `Drop`. That way, there's no risk of forgetting to do the transfer for each model.

An alternative suggestion would be to create a wrapper `OffloadableTensor(ggml::Tensor)`, which does the transfer back on `Drop`. That way, you can explicitly choose to opt into that behaviour for model tensors.

crates/models/llama/src/lib.rs
```diff
@@ -52,39 +63,54 @@ impl KnownModel for Llama {
 
         // model-global weights
         let wte = tl.load("tok_embeddings.weight")?;
-        let norm = tl.load("norm.weight")?;
-        let output = tl.load("output.weight")?;
+
+        let backend = if params.should_offload(0) {
```
philpax, line 67: Seems like we should have a more explicit way of checking this. Maybe a `params.backend()` that returns the correct backend?

crates/models/llama/src/lib.rs
```diff
@@ -52,39 +63,54 @@ impl KnownModel for Llama {
...
+            Backend::Gpu
+        } else {
+            Backend::Cpu
+        };
+
+        let norm = tl.offload("norm.weight", backend)?;
```
philpax, line 73: Ah, I see, the `offload` is a bit more convenient. I think `load("norm.weight")?.transfer_to(backend)?` is an acceptable compromise to reduce the complexity of the loader. (PyTorch uses `to`, which could also work, but I'm not sure if that's too short for Rust.)

crates/models/llama/src/lib.rs
```diff
@@ -52,39 +63,54 @@ impl KnownModel for Llama {
...
+
+        let output = tl.offload("output.weight", backend)?;
 
         let mut layers = Vec::new();
+
         for i in 0..hyperparameters.n_layer {
+            let backend = if params.should_offload(i) {
```
philpax, line 80: Okay, maybe `params.backend(layer_index: usize)`.

As an aside - because `should_offload(0)` is used for both `norm/output`/the rest of the layer, it means that layer 0 will be larger than the others. Not sure if it matters too much.

crates/models/llama/src/lib.rs
```diff
@@ -121,25 +145,26 @@ impl KnownModel for Llama {
...
-                    1
-                } else {
-                    num_threads
-                },
-            );
+            let mut gf = ggml::ComputationGraph::new();
+
             for il in 0..n_layer {
+                //TODO: find a better way to do this
+                if self.model_params.should_offload(il) {
```
philpax, line 158: `ctx0.set_offloading_enabled(self.model_params.should_offload(il))`
or
`ctx0.set_backend(self.model_params.backend(il))`

philpax (changes requested):
Really impressive work! I'm very thankful you've taken the lead on this.

I still need to test it personally (haven't been around my Windows machine much in the last two days), but this is looking good. I've made a few comments about style/minor tweaks, but the core of this looks excellent.

Looking forward to seeing how quickly my GPU can run LLaMA 🚀

LLukas22:
Thanks for the review, i'll try to implement the changes later today. As previously mentioned some models currently produce gibberish when Cuda acceleration is enabled, thats something i also have to look into.

crates/models/llama/src/lib.rs
LLukas22: `params.backend(layer_index: usize)` Sounds like a good idea. `llama.cpp` handles the `norm/output` by checking if the number of layers that should be offloaded is >= the number of layers in the model. If it is they will be offloaded. I decided to offload them if at least one layer is offloaded, maybe we should implement it similar to `llama.cpp` to avoid unnecessary copying from ram->gpu->ram->gpu 🤔

crates/models/llama/src/lib.rs
LLukas22: `to` could be confusing as the offloading could change the underlying tensor structure, which `to` in pytorch doesn't do.
`transfer_to` seams fine.

crates/models/llama/src/lib.rs
LLukas22: I was thinking about moving this into the `ggml::context` as an optional `loaded_tensors` field, which is populated by the loader. As far as im aware  each model keeps its own `ggml::context` which means the context is dopped if the model is droppen. Meaning we just have to free the `loaded_tensors` when a context is dropped.

crates/llm-base/src/loader.rs
LLukas22: Was 32, now is 48. I will use `GGML_MAX_NAME`.

crates/models/llama/src/lib.rs
philpax: Honestly it's a little surprising that the context doesn't clean this up itself when it's free.

I think having the context be mutated by the loader might be a bit weird in terms of cross-cutting concerns. Maybe have `Tensor::transfer_to` update the context's internal list of offloaded tensors (making sure to remove a tensor from that list if it's back on the CPU), and then having the context clean up that list on `Drop`?

crates/ggml/src/context.rs
LLukas22: Can we change this to `set_offloading(&mut self, can_offload: bool)`? The we only need one function and i would like to keep the boolean field private 🤔

crates/llm/examples/drop_multiple_sessions.rs
LLukas22: I'll accidentally committed this. Will remove it before merging.

crates/llm/examples/drop_multiple_sessions.rs
LLukas22: Integration testing this is not possible as in a cpu only setting the context will always cleanup the tensors. The explicit free is only needed when opencl or cuda is used, which the github actions runners don't support.

crates/llm-base/src/lora.rs
LLukas22: Good idea but i probably wont implement this in this PR 👍

crates/models/llama/src/lib.rs
LLukas22: Solved it with some horrible `Arc<Mutex<HashMap>>` thingy. There probably is a better way of doing this 😓

crates/llm-base/src/inference_session.rs
```diff
@@ -238,7 +220,7 @@ impl InferenceSession {
         self.ctx0 = ggml::Context::init_buffer(self.ctx0.buffer.take().unwrap());
         let ctx0 = &mut self.ctx0;
         let mut embd = ctx0.new_tensor_1d(ggml::Type::I32, input_tokens.len());
-        ggml::set_name(&embd, "embd");
+        ggml::set_tensor_name(&embd, "embd");
```
philpax, line 223: Any particular reason not to use `embd.set_name("embd")`?

crates/llm-base/src/inference_session.rs
```diff
@@ -946,3 +928,24 @@ pub fn feed_prompt_callback<'a, E: std::error::Error + 'static>(
...
+    let memory_k = context
+        .new_tensor_1d(config.memory_k_type.into(), n_elements)
+        .set_name("memory_k");
+    let memory_v = context
+        .new_tensor_1d(config.memory_v_type.into(), n_elements)
+        .set_name("memory_v");
+
+    if config.use_gpu {
+        ggml::accelerator_offload_tensor_no_scratch(&memory_k);
+        ggml::accelerator_offload_tensor_no_scratch(&memory_v);
```
philpax, line 947: Should this be part of the `Tensor` interface too?

crates/models/llama/src/lib.rs
```diff
@@ -64,39 +53,49 @@ impl KnownModel for Llama {
...
-                wk: tl.offload(&format!("layers.{i}.attention.wk.weight"), backend)?,
-                wv: tl.offload(&format!("layers.{i}.attention.wv.weight"), backend)?,
-                wo: tl.offload(&format!("layers.{i}.attention.wo.weight"), backend)?,
-                ffn_norm: tl.offload(&format!("layers.{i}.ffn_norm.weight"), backend)?,
-                w1: tl.offload(&format!("layers.{i}.feed_forward.w1.weight"), backend)?,
-                w2: tl.offload(&format!("layers.{i}.feed_forward.w2.weight"), backend)?,
-                w3: tl.offload(&format!("layers.{i}.feed_forward.w3.weight"), backend)?,
+                    .load(&format!("layers.{i}.attention_norm.weight"))?
+                    .transfer_to(backend)?,
+                wq: tl
```
philpax, line 71: hm okay, I was wrong, this is a bit unwieldy with the `transfer_to` calls. Can you try adding an `Option<Backend>` parameter to `load` and see how that looks?

crates/models/llama/src/lib.rs
```diff
@@ -64,39 +53,49 @@ impl KnownModel for Llama {
...
+                    .load(&format!("layers.{i}.feed_forward.w2.weight"))?
+                    .transfer_to(backend)?,
+                w3: tl
+                    .load(&format!("layers.{i}.feed_forward.w3.weight"))?
+                    .transfer_to(backend)?,
             };
             layers.push(layer);
         }
-        let (context, loaded_tensors) = tl.finish();
+        let (context, _) = tl.finish();
```
philpax, line 98: It doesn't look like we actually use the loaded tensors in any of the models; maybe we should just remove the second return value, so that it can be `let context = tl.finish();`

crates/ggml/src/tensor.rs
```diff
@@ -27,29 +42,60 @@ impl Tensor {
...
     }
 
-    ///Gets the name of the tensor
-    pub fn get_name(&self) -> String {
+    /// Gets the name of the tensor
+    pub fn name(&self) -> String {
         let name = unsafe { self.ptr.as_ref().name };
         let mut name = name.iter().map(|&x| x as u8).collect::<Vec<_>>();
         name.retain(|&x| x != 0);
         String::from_utf8(name).unwrap()
```
philpax, line 50: Okay, this one's not on you - I should have caught it, and if you don't fix it I will later - but this should probably use `CString::new` and `to_string_lossy`, because this isn't guaranteed to be UTF-8 (although 99.9999% of the time it will be), and `CString` can better maintain the invariants

crates/models/llama/src/lib.rs
```diff
@@ -185,7 +178,7 @@ impl KnownModel for Llama {
                     n_rot,
                     0,
                 );
-                ggml::set_name(&q_current, "Qcur");
+                ggml::set_tensor_name(&q_current, "Qcur");
```
philpax, line 181: ah I see, these calls were there before. We should make `set_tensor_name` private to `ggml`, and force everyone to use `Tensor::set_name`.

philpax (changes requested):
Almost there...

crates/ggml/src/context.rs
```diff
@@ -21,6 +26,9 @@ pub struct Context {
 
     /// Whether the context can offload tensors to the GPU
     pub can_offload: bool,
+
+    /// Offloaded tensors
+    offloaded_tensors: Arc<Mutex<HashMap<String, Tensor>>>,
```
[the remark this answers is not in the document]
LLukas22: This is flawed either way, Tensors aren't guaranteed to have a name, except they were created by the loader. We shouldn't overcomplicate this as there is a high chance, that upstream will change the context to automatically cleanup these offloaded tensors.

crates/ggml/src/tensor.rs
```diff
@@ -27,29 +42,60 @@ impl Tensor {
...
     }
 
+    /// Sets the tensors acceleration backend and moves the tensors data to the new backend.
+    pub fn transfer_to<E>(mut self, backend: crate::Backend) -> Result<Tensor, E> {
+        let current_backend = self.backend();
+        self.set_backend(backend);
+
+        if backend != crate::Backend::Cpu {
+            crate::accelerator_transform_tensor(&mut self);
+            if current_backend == crate::Backend::Cpu {
```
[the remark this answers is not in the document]
LLukas22: Currently there is no way to manually move a tensor back to the CPU. That's why i ignored that case.

crates/llm-base/src/inference_session.rs
LLukas22: This is very specific for CUDA and the K/V-memory. No other Tensors should be offloaded without scratch.

crates/ggml/src/context.rs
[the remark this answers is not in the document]
philpax: Fair enough, keep it as-is then

crates/ggml/src/tensor.rs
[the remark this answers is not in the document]
philpax: Ah, fair enough. Can you add an assert that the transfer target isn't the CPU?

crates/llm-base/src/inference_session.rs
philpax: In that case, add a comment to that effect so that we remember the reasoning for this.

crates/models/llama/src/lib.rs
LLukas22: I don't know if adding a `Option<Backend>` to `load` would be better. As we still have to always define a value for it. With the `Tensor::transfer_to`  implementation the offloading is completely decoupled from the loader which i quite like. But i don't know how we could make this easier to use 🤔

crates/ggml/src/tensor.rs
```diff
@@ -27,29 +42,60 @@ impl Tensor {
...
+                // tensor was moved from cpu to accelerator => We need to keep track of the data to free it later from the accelerator
+                self.with_alive_ctx_mut(|| {
+                    if let Some(offloaded_tensors) = self.offloaded_tensors.upgrade() {
+                        //TODO: Do we need to check if the tensor is already in the map?
+                        offloaded_tensors
```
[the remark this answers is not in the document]
LLukas22: Could you handle the `ContextInner` thing?

crates/models/llama/src/lib.rs
philpax: Yeah, good point. Maybe, instead of having individual calls, you instead have

`for tensor in &[...] { tensor.transfer_to(backend)?; }`

Another alternative would be defining `transfer_to` on `Layer` itself, so that you can transfer the entire layer as a unit without worrying about mixing loading and transfer logic.

crates/models/llama/src/lib.rs
LLukas22: Should we change this in this PR or create another one?

philpax: I'll create another one after this to fix up the names. Don't worry about it.

crates/models/llama/src/lib.rs
LLukas22: I don't know if that's better. If we do the `transfer_to` on a `Layer` we simply hide the `transfer_to` calls of each tensor behind another function. And we still have to implement it for each model as they use different `Layer` structs.

We also have to think about the multi GPU case. There we can offload certain tensors of a layer to all GPUs but some maybe need to be synced to a single GPU 🤔

crates/ggml/src/tensor.rs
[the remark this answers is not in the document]
philpax: Yeah, sure, just let me know when you're ready for me to make commits

crates/models/llama/src/lib.rs
philpax: Alright, yeah. Leave it as-is for now then, we'll revisit once we have other models and/or multi-GPU support.

crates/models/llama/src/lib.rs
LLukas22: I guess this also is another cleanup task? I actually don't know why `finish` returns the tensors 😬

philpax: I'm sure there was a good reason at some point, but we don't need it any more. Up to you if you want to clean it up or not; I can do that otherwise.

LLukas22:
@philpax I closed all conversations i fixed in my last commit. Would be great if you could take a look at the rest.

philpax:
Nice work! I'll test it locally, make any final changes of my own, and then I'll merge it 🚀 

(Don't worry about solving the merge conflicts, I'll do that myself)

philpax:
I'm going to sleep on this before I merge it. I think my changes should all check out, but it's quite late now and I'm pretty sure the GPU offloading stuff did a number on my brain.

Pending issues (that will be made into issues after this is merged):
- Certain quantization levels do not work.
    - 7B Q5_1 does not work (Philpax)
    - 7B Q3_K_M works (Philpax)
    - 13B Q4_K_M works (Philpax)
    - 13B Q5_K_M works (Lukas)
    - 13B Q5_K_S does not work (Lukas)
- ExecutionParameters struct for passing thread count and backend around

crates/ggml/src/tensor.rs
```diff
@@ -15,27 +17,110 @@ impl Tensor {
...
+    /// It will not transfer the data. Use `transfer_to` for that.
+    #[allow(unused_variables)]
+    pub fn offload(&self) {
+        self.with_alive_ctx(|| {
+            #[cfg(feature = "cublas")]
+            unsafe {
+                sys::cuda::ggml_cuda_assign_buffers(self.ptr.as_ptr());
+            }
+        })
     }
```
LLukas22, line 99: Does this automatically set the correct backend? And how would it behave in a multi-GPU setting?

crates/ggml/src/tensor.rs
```diff
@@ -15,27 +17,110 @@ impl Tensor {
...
+
+    /// Sets the tensor's acceleration backend and moves the tensor's data to the new backend.
+    pub fn transfer_to(mut self, backend: Backend) -> Tensor {
+        self.with_alive_ctx_mut(|t| {
+            let current_backend = t.backend();
+
+            if current_backend != Backend::Cpu && backend == Backend::Cpu {
+                unimplemented!("Tensors cannot be moved from an accelerator to the CPU at present");
+            }
+            t.set_backend(backend);
```
LLukas22, line 68: Shouldn't `t.set_backend(backend)` happen after the `if` block?

crates/ggml/src/tensor.rs
philpax: > Does this automatically set the correct backend?

I think so: 
```c
void ggml_cuda_assign_buffers_impl(struct ggml_tensor * tensor, bool scratch, bool force_inplace) {
    /* .. */

    tensor->backend = GGML_BACKEND_GPU;
}
```

> And how would it behave in a multi-GPU setting?

Great question. It appears that it always ends up on the main GPU and it's moved by `transform_tensor`. I guess this means that the model is the main thing distributed between GPUs.

LLukas22: Seams correct, does this new `offload` method also work with opencl? I thought we also have to set the backend to `GPU` for all computation tensors if we use opencl🤔

philpax: Not sure; this should be doing the same thing as the previous `ggml::accelerator_offload_tensor`, so if it was working before, I want to assume it'll keep working?

Kind of annoying to test though :/

LLukas22: Just tested OpenCl and it still works 👍

PR #326, opened 2023-06-22
K-quants

### LINKED ISSUE #301: Add support for new k-quants quantization format
`llama.cpp` now supports new k-quants quantizations which achieve good model perplexity even in high quantizations. See  https://github.com/ggerganov/llama.cpp/pull/1684 .

We should also support these new quantization formats.

nightscape:
This is the first baby steps. I'm mechanically adding the `QX_K` everywhere where the other quantizations are listed.
I'd be very thankful for early feedback :smiley:
This fixes #301


binaries/generate-ggml-bindings/src/main.rs
```diff
@@ -12,8 +12,8 @@ fn main() {
     let src_path = sys_path.join("src");
 
     let bindings = bindgen::Builder::default()
-        .header(ggml_path.join("ggml.h").to_str().unwrap().to_string())
-        .allowlist_file(r".*ggml.h")
+        .header(ggml_path.join("k_quants.h").to_str().unwrap().to_string())
```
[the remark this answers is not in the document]
LLukas22: I actually don't know. If you generate the [bindings](https://github.com/rustformers/llm/blob/main/doc/CONTRIBUTING.md#regenerating-ggml-bindings) the functions listed in `k_quants.h` should be automatically wrapped inside the generated `ggml\sys\src\lib.rs` file. If thats the case then it should be correct.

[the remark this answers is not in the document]
nightscape: It seems to work at least there are no compile errors.

nightscape:
After the last fixup it compiles, loads the tensors (according to the console output) and even generates text that looks good!!
```
     Running `target/release/llm llama infer --use-gpu --model-path /Users/martin/Downloads/models/airoboros-7b-gpt4-1.4.ggmlv3.q2_K.bin --prompt 'Once upon a time' --no-mmap`
✓ Loaded 291 tensors (2.9 GB) after 795ms
ggml_metal_init: allocating
ggml_metal_init: using MPS
ggml_metal_init: loading '/Users/martin/Workspaces/ai/llm/target/release/ggml-metal.metal'
ggml_metal_init: loaded kernel_add                            0x143628f20
ggml_metal_init: loaded kernel_mul                            0x143632550
ggml_metal_init: loaded kernel_mul_row                        0x143631bb0
ggml_metal_init: loaded kernel_scale                          0x143633cb0
ggml_metal_init: loaded kernel_silu                           0x1436346e0
ggml_metal_init: loaded kernel_relu                           0x1436333a0
ggml_metal_init: loaded kernel_gelu                           0x143635c30
ggml_metal_init: loaded kernel_soft_max                       0x14370c5a0
ggml_metal_init: loaded kernel_diag_mask_inf                  0x14370ce20
ggml_metal_init: loaded kernel_get_rows_f16                   0x14370d090
ggml_metal_init: loaded kernel_get_rows_q4_0                  0x1336043e0
ggml_metal_init: loaded kernel_get_rows_q4_1                  0x1336050a0
ggml_metal_init: loaded kernel_get_rows_q2_k                  0x133605940
ggml_metal_init: loaded kernel_get_rows_q3_k                  0x1336060e0
ggml_metal_init: loaded kernel_get_rows_q4_k                  0x1336079d0
ggml_metal_init: loaded kernel_get_rows_q5_k                  0x1336071e0
ggml_metal_init: loaded kernel_get_rows_q6_k                  0x1336089f0
ggml_metal_init: loaded kernel_rms_norm                       0x133606b50
ggml_metal_init: loaded kernel_norm                           0x1336093c0
ggml_metal_init: loaded kernel_mul_mat_f16_f32                0x13360a240
ggml_metal_init: loaded kernel_mul_mat_q4_0_f32               0x13360a4f0
ggml_metal_init: loaded kernel_mul_mat_q4_1_f32               0x13360ad00
ggml_metal_init: loaded kernel_mul_mat_q2_k_f32               0x13360b5b0
ggml_metal_init: loaded kernel_mul_mat_q3_k_f32               0x13360be20
ggml_metal_init: loaded kernel_mul_mat_q4_k_f32               0x13360c6a0
ggml_metal_init: loaded kernel_mul_mat_q5_k_f32               0x1440fda60
ggml_metal_init: loaded kernel_mul_mat_q6_k_f32               0x1440ffcb0
ggml_metal_init: loaded kernel_rope                           0x1440fe1a0
ggml_metal_init: loaded kernel_alibi_f32                      0x1440fe840
ggml_metal_init: loaded kernel_cpy_f32_f16                    0x123f06f20
ggml_metal_init: recommendedMaxWorkingSetSize = 10922.67 MB
ggml_metal_init: hasUnifiedMemory             = true
ggml_metal_init: maxTransferRate              = built-in GPU
ggml_metal_add_buffer: allocated 'scratch         ' buffer, size =  1024.00 MB, ( 1024.39 / 10922.67)
ggml_metal_add_buffer: allocated 'scratch         ' buffer, size =   512.00 MB, ( 1536.39 / 10922.67)
ggml_metal_add_buffer: allocated 'scratch         ' buffer, size =   512.00 MB, ( 2048.39 / 10922.67)
ggml_metal_add_buffer: allocated 'wt              ' buffer, size =  1024.09 MB, ( 3072.48 / 10922.67)
ggml_metal_add_buffer: allocated 'wt              ' buffer, size =  2733.66 MB, ( 5806.14 / 10922.67)
Once upon a time, there was an extraordinary woman.
Her name is Sara-Lena. She has been on this Earth for over twenty years now and she has done so many things in her life that most people would be afraid to do or even think about doing them because of their fears, but not her! Her passionate heart led her into experiences which others dreamt about while they were still alive; nevertheless.
She is a spiritually grounded being who knows herself well and loves everything she does in life - like dancing barefoot on the beach at sunrise or writing love letters to trees that have seen so much more than we ever could... She has been many things, but now her path leads towards empowering others through words and pictures.
With a background as an actress, model, writer/poetess (and soon-to-be author of two books!), yogi teacher & energy healer she is here to share the lessons learned on this wild journey called life - so stay tuned for more stories from Sara's Heart.
ggml_metal_free: deallocating
```

I have no idea if this is correct though... I would have expected it to be much harder 😉 
Could you maybe have a look if this could be sufficient already @LLukas22 @philpax @pixelspark ?

crates/ggml/sys/build.rs
```diff
@@ -180,6 +183,24 @@ fn enable_metal(build: &mut cc::Build) {
...
+        .unwrap()
+        .join("target")
+        .join(build_type);
+    let output_path = PathBuf::from(path);
+
+    let input_path =
+        Path::new(&env::var("CARGO_MANIFEST_DIR").unwrap()).join("llama-cpp/ggml-metal.metal");
+    let output_path = Path::new(&output_path).join("ggml-metal.metal");
+    std::fs::copy(input_path, output_path).unwrap();
+
```
LLukas22, line 203: This should not be part of this PR. As you already opened another PR for this change 😅

crates/llm-base/src/loader.rs
```diff
@@ -57,6 +62,7 @@ impl TryFrom<i32> for FileType {
             7 => FileTypeFormat::MostlyQ8_0,
             8 => FileTypeFormat::MostlyQ5_0,
             9 => FileTypeFormat::MostlyQ5_1,
+            10 => FileTypeFormat::MostlyQ2_K,
```
LLukas22, line 65: Aren't the other k_quants missing here?

nightscape: There is a `Q8_K`, but it is defined [in some places](https://github.com/ggerganov/llama.cpp/blob/master/ggml.h#L249), but [not in others](https://github.com/ggerganov/llama.cpp/blob/master/ggml.h#L278).

crates/ggml/sys/build.rs
nightscape: 👍  I rebased and took that commit out.

crates/llm-base/src/loader.rs
```diff
@@ -40,6 +40,11 @@ impl From<FileType> for i32 {
                 FileTypeFormat::MostlyQ8_0 => 7,
                 FileTypeFormat::MostlyQ5_0 => 8,
                 FileTypeFormat::MostlyQ5_1 => 9,
```
nightscape, line 42: @philpax is there a reason for not reusing the constants defined in `ggml.h` here?

crates/llm-base/src/loader.rs
LLukas22: Aren't `Q3_K` to `Q6_K` missing here? you implemented them in `From<FileType>` function but not in the `TryFrom<i32>` function.

crates/llm-base/src/loader.rs
LLukas22: I think the reason was convenience, we should swap them with the constants, similarly on how it's done on the [Type](https://github.com/rustformers/llm/blob/main/crates/ggml/src/lib.rs#L126C12-L187) enum.

crates/llm-base/src/loader.rs
nightscape: Right 👍
Fixed!

crates/llm-base/src/loader.rs
nightscape: I think I see now why it was not trivial:
The corresponding constants are defined in [`crates/ggml/sys`](https://github.com/rustformers/llm/tree/main/crates/ggml/sys), which is a nested crate inside [`crates/ggml`](https://github.com/rustformers/llm/tree/main/crates/ggml).
At least according to GPT-4, one would have to add `crates/ggml/sys` as a direct dependency to `llm-base`, which might not be desired?

nightscape:
Would it be ok to merge this PR as is and create issues for the quantization and saveguards?
I might be running out of time and skills at any point and then the (apparently) already working k-quant reading would be delayed.

pixelspark:
I have very little knowledge on the quantization stuff, so unfortunately little to add here - @nightscape your suggestion to just merge this in a working state sounds good.

I don't see anything obviously wrong with the output shown above though. Perhaps it is a good idea to nevertheless do a quick compare between `llama.cpp` itself and your test runs.

nightscape:
@pixelspark you mean by setting the same seed in both projects and checking if it generates the same text?

nightscape:
Hmm, that doesn't seem to work. I tried to make all parameters identical, but don't get the same text.
Basically all calls to the random generator would have to be done in the exact same sequence, which is rather improbable.
@pixelspark I guess you meant something else, right?

nightscape:
Looks like `q4_K_M` models don't work yet, so I'll need another iteration.

crates/llm-base/src/loader.rs
LLukas22: Maybe we could move the enums into `crates/ggml` to avoid this?

pixelspark:
> Hmm, that doesn't seem to work. I tried to make all parameters identical, but don't get the same text. Basically all calls to the random generator would have to be done in the exact same sequence, which is rather improbable. @pixelspark I guess you meant something else, right?
> 
> The generated texts all look sensible I can't see a difference in the quality of generated text.

I just meant a superficial comparison (if you can do a fully deterministic comparison that would be better but as you say, that is quite difficult to do). Good to hear it looks about the same :-)

crates/llm-base/src/loader.rs
philpax: That would be a great idea and was what I was going to do (exposing the `sys` crate so that the enum could be defined in terms of those values), but...

they're actually different in `llama.cpp` 🤦 

https://github.com/ggerganov/llama.cpp/blob/7487137227eb32ed9b12156338b865cb29b2dfd1/llama.h#L96-L116

That also explains why q4_k_m doesn't work right now. Let me see if I can work around this somehow.

philpax:
Alrighty, this should be good to go and support all of the ftypes! Thanks so much for getting the ball rolling on this 🚀 

If you're happy with this and it passes all of the CI, I'll merge it in tomorrow.

PR #328, opened 2023-06-22
fix(metal): patch ggml-metal.m with shader code

philpax:
Alternate fix to #324 (thanks for the inspiration!)

The Rust build script has the ability to create new files, and use them as fodder for the build process. We can use this to create an alternate `ggml-metal.m` that embeds the Metal shader source code and sidestep the issue of runtime loading entirely.

Let me know if you're OK with this @pixelspark @nightscape !

### DIFF
--- a/crates/ggml/sys/build.rs
+++ b/crates/ggml/sys/build.rs
@@ -201,10 +201,24 @@ fn enable_metal(build: &mut cc::Build, out_dir: &Path) {
             panic!("ggml-metal.m does not contain the needle to be replaced; the patching logic needs to be reinvestigated. Contact a `llm` developer!");
         }
 
+        // Replace the runtime read of the file with a compile-time string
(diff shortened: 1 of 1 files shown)


philpax:
Also took the liberty of patching the calls to `printf` so that they wouldn't spam output in normal release mode. Happy to revert that commit if we think it's too much.

nightscape:
Looks great!!

pixelspark:
Nice trick, maybe a bit brittle (although I suspect the line you replace is not going to change any time soon). 

A cleaner way would be to make a trivial PR to llama.cpp that would make the code accept a GGML_METAL_SHADER_CODE #define and use that when set, or else loads from ggml-metal.m? (Assuming the length of #define contents is not limited. Otherwise we could add a #define that chooses between load-at-runtime or a #include).

I propose to merge this now but also initiate steps for a ‘clean’ version.

PR #337, opened 2023-06-30
Basic Integration Testing

LLukas22:
Starting point for https://github.com/rustformers/llm/issues/319

Tests can be run via `cargo run --release --bin llm-test` and they are integrated into CI. Tests for a specific architecture can be run by defining the architecture in the cli call e.g. `cargo run --release --bin llm-test llama`. If the models needed for testing can't be found on your system they will be downloaded from huggingface. 

Currently they only test if a model is loadable and can generate some tokens. The actual output is not validated.


philpax:
Cursory look through looks good, I'll pull it down, test it, and get it merged later today or tomorrow 💪

philpax:
I've updated the PR with some clean-up commits, I'll write up the tests later today (will probably put some test cases in the configs) and then this should be good to go

philpax:
Alrighty, this should be good to go. @LLukas22 can you check over my changes? This should let us add more tests as required.

LLukas22:
Looks good, but shouldn't the "can send" test be test case?

philpax:
I thought about that, but I figured that it didn't make sense to make a one-line check an entirely separate type of test case. I'll think about it and see if there's anything better we can do 🤔 

Thanks for getting this started and checking over my changes! I'm super glad that we now have integration tests, especially with all of the architectures! Awesome work 💪

PR #343, opened 2023-07-02
fix #338 - use wte if no lm_head for gpt2

### LINKED ISSUE #338: GPT-2 doesn't always have an `lm_head`
The GPT2 ggml models don't always have an lm_head, but `llm` assumes it does:

https://github.com/rustformers/llm/blob/9a222690cd0a9e3322bb6a926d54e90d08fb2c0f/crates/models/gpt2/src/lib.rs#L62

See ggml example repo where they mention lm_head is optional: https://github.com/ggerganov/ggml/blob/ee1b3727e60403012dd2b57d35b60558f4db66d8/examples/gpt-2/main.cpp#L364

philpax:
I think this should fix #338, but I don't have a model without a LM head to test.

@steventrouble do you have a sample model to test with?

### DIFF
--- a/crates/models/gpt2/src/lib.rs
+++ b/crates/models/gpt2/src/lib.rs
@@ -30,8 +30,10 @@ pub struct Gpt2 {
     wte: Tensor,
     // weighted positional encodings
     wpe: Tensor,
-    // language model head
-    lm_head: Tensor,
+    // language model head.
+    //
+    // Optional: if not present, the `wte` tensor is used instead.
+    lm_head: Option<Tensor>,
 
     // weights for the model
     layers: Vec<Layer>,
@@ -59,7 +61,9 @@ impl KnownModel for Gpt2 {
         let ln_f_b = tl.load("model/ln_f/b")?;
         let wte = tl.load("model/wte")?;
         let wpe = tl.load("model/wpe")?;
-        let lm_head = tl.load("model/lm_head")?;
+        // GPT-2's language model head is optional; if it is not present,
+        // the `wte` tensor is used instead.
+        let lm_head = tl.load("model/lm_head").ok();
 
         let mut layers = Vec::new();
         for i in 0..hyperparameters.n_layer {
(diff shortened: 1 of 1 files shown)


LLukas22:
See ggerganov's original [gpt-2 models](https://huggingface.co/ggerganov/ggml/tree/main).

philpax:
Hm, ok, tested with that - turns out it segfaults due to an out-of-bounds memory access. Will investigate some other time...

steventrouble:
I found the node that's causing the segfault. It's the 1d view of memory_k [here](https://github.com/steventrouble/llm/commit/00fd813e5f8e40073afe81bc47047ed171ffaec2#diff-897ee2a80b3fd6a96150b2579bebddf1e7349d9a64e6db55a9ce16ab282f1382R177) being called with an out-of-bounds offset.

[Here's an assertion](https://github.com/steventrouble/llm/commit/600f7c74472871b89af2a3a8865f1df9acc93ed3#diff-afd9229eeb711b9449693886fe0cf6c7946661ae969bf9d1409ec6406eedc779R281) that may help you repro.

LLukas22:
I'll try to take a look at this later but good job finding this, saves me a lot of time 👍

philpax:
> I found the node that's causing the segfault. It's the 1d view of memory_k [here](https://github.com/steventrouble/llm/commit/00fd813e5f8e40073afe81bc47047ed171ffaec2#diff-897ee2a80b3fd6a96150b2579bebddf1e7349d9a64e6db55a9ce16ab282f1382R177) being called with an out-of-bounds offset.
> 
> [Here's an assertion](https://github.com/steventrouble/llm/commit/600f7c74472871b89af2a3a8865f1df9acc93ed3#diff-afd9229eeb711b9449693886fe0cf6c7946661ae969bf9d1409ec6406eedc779R281) that may help you repro.
> 
> The issue seems related to the similarity between the names `hyperparameters.n_ctx` and `context_size`. Updating `start_session` to use `self.context_size` instead of `self.hyperparameters.n_ctx` fixes the issue for me, but I don't know enough about ggml to be sure that's the root cause. [Example commit](https://github.com/steventrouble/llm/commit/52eaa6fcacb0ce5ed77d19077b3fd12848f714f8)

Well spotted! Feel free to open a PR with that and I'll close this. What you've (correctly) identified there is that that the context length that the model supports (hyperparameters n_ctx) is not the same as the context length that's actually in use. Oops.

philpax:
Obsoleted by #362

PR #351, opened 2023-07-04
Add placeholder check for message prompt

viirya:
When playing with chat command and just found that it doesn't check if provided message prompt and message prompt contain correct placeholder. If it is not, no proper feedback to users and users won't know the prompt isn't actually generated to the model.


philpax:
Thanks!

viirya:
Thank you @LLukas22 @philpax !

PR #360, opened 2023-07-09
Add CI for cuda builds

LLukas22:
Adds cublas builds to CI. 

Currently the cuda-toolkit setup on a windows runner takes ~15 min (see https://github.com/Jimver/cuda-toolkit/issues/253), maybe we should disable the the windows cuda ci for the time being?


philpax (approved):
I'm fine with the Windows build taking a while, merge away when you're comfortable

PR #362, opened 2023-07-09
Update gpt2 to use wte if no lm_head

### LINKED ISSUE #338: GPT-2 doesn't always have an `lm_head`

steventrouble:
Closes #338 

Based off of https://github.com/rustformers/llm/pull/343

Fixes the segfault issue mentioned in that bug, and adds a check that could help catch those errors faster.

### DIFF
--- a/crates/ggml/src/context.rs
+++ b/crates/ggml/src/context.rs
@@ -278,6 +278,11 @@ impl Context {
 
     /// Creates a 1D view over `a`.
     pub fn op_view_1d(&self, a: &Tensor, ne0: usize, offset: usize) -> Tensor {
+        #[cfg(debug_assertions)]
+        assert!(
+            offset < a.nbytes(),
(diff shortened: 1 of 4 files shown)


philpax:
Looks good! Can you add the comments from the original PR about why the tensor's optional and why we substitute with `wte`?

steventrouble:
👍 done, thanks!

philpax:
Brilliant, thanks 🚀

PR #364, opened 2023-07-09
allow chat to halt new token generation on `stop_sequence`

averypelle:
Closes https://github.com/rustformers/llm/issues/363

Stop token generation after reaching a specified `stop_sequence` in chat mode

I am still new to rust so please let me know how I can improve my code!


binaries/llm-cli/src/main.rs
```diff
@@ -264,6 +266,33 @@ fn interactive(
             .unwrap_or(false)
     }
 
+    fn inference_callback(
```
averypelle, line 269: this is essentially a copy of the function here https://github.com/rustformers/llm/blob/main/crates/llm/examples/vicuna-chat.rs#L119

binaries/llm-cli/src/main.rs
```diff
@@ -256,6 +256,12 @@ fn interactive(
     let parameters = generate.inference_parameters(model.eot_token_id());
     let mut rng = generate.rng();
 
+    let stop_sequence = message_prompt_template
```
averypelle, line 259: I am assuming the message_prompt_template is something like:
```
User: {{PROMPT}}
```

But very open to suggestions here

binaries/llm-cli/src/main.rs
```diff
@@ -448,3 +446,42 @@ impl Validator for LineContinuationValidator {
...
 }
+
+fn inference_callback(
+    stop_sequence: String,
+    chat_mode: bool,
+    buf: &mut String,
+) -> impl FnMut(InferenceResponse) -> Result<InferenceFeedback, Infallible> + '_ {
+    move |resp| match resp {
+        InferenceResponse::InferredToken(t) => {
+            if chat_mode {
```
averypelle, line 457: i didn't touch REPL mode - is the desired behavior to continue generating tokens in that mode? otherwise happy to change

philpax:
Great work! I was actually thinking about bringing that logic out so that I could use it for `llmcord`.

Do you think you'd be able to move the `inference_callback` to `llm-base`, name it something like `conversation_inference_callback`, and update both `llm-cli` and the `vicuna-chat` example to use it? (You might need to parameterise over `print_token`)

That would allow for this logic to be used across both, as well as elsewhere (the aforementioned `llmcord`).

I'd suggest passing the stop sequence in from the CLI (i.e. maybe replace `message_prompt` with `message_prompt_prefix` and use that as the stop sequence.)

> i didn't touch REPL mode - is the desired behavior to continue generating tokens in that mode? otherwise happy to change

Yup, that's for a back and forth where no state is preserved and the model can produce as much output as it wants. I'd suggest splitting the code paths so that they use entirely different inference callbacks - they share the readline, but their inference behaviour is pretty different.

Great work once again, let me know if you need a hand with any of this! 🙂

averypelle:
Thanks @philpax! Just pushed an update where I moved the function to `llm-base`. For the `message_prompt`, is there any case where someone would want a template that includes a postfix? If so, maybe a new option is needed? Otherwise, I can rename to prefix - in this case, would it still make sense for the prefix to include the string `{{PROMPT}}`?

crates/llm-base/src/inference_session.rs
```diff
@@ -853,3 +853,36 @@ pub fn feed_prompt_callback<'a, E: std::error::Error + 'static>(
         None => Ok(InferenceFeedback::Continue),
     }
 }
+
+/// Callback to be passed to [InferenceSession::infer] that will print the
```
philpax, line 857: Can you update this comment to better describe what the function does/how it's different to the normal callback?

binaries/llm-cli/src/main.rs
```diff
@@ -283,26 +297,40 @@ fn interactive(
...
+                },
+                &mut Default::default(),
+                conversation_inference_callback(stop_sequence.clone(), &mut buf, print_token),
+            )
+        } else {
+            session.infer::<Infallible>(
+                model.as_ref(),
+                &mut rng,
+                &llm::InferenceRequest {
+                    prompt: "".into(),
```
philpax, line 318: Are the prompts meant to be empty here?

philpax:
Great work! Looking forward to merging this soon 🚀 

> For the `message_prompt`, is there any case where someone would want a template that includes a postfix? If so, maybe a new option is needed? Otherwise, I can rename to prefix - in this case, would it still make sense for the prefix to include the string `{{PROMPT}}`?

I don't think the logic would work if it was postfix anyway - we should make it clear that you *need* to pass in a prefix. I'd say you can leave out the `{{PROMPT}}` in that case, because it should always be implied that the prompt will be suffixed.

binaries/llm-cli/src/main.rs
averypelle: ooh great find! i don't suspect so. this was in the previous impl, let me try with `prompt: prompt.as_str().into(),` and test out

averypelle: no that's not it. I think it needs to be `message_prompt_template` or `initial_prompt_template` depending on the call number. But with that kind of branching logic, it might be time to more fully separate these code paths

philpax: Oh, disregard, I'm a silly billy - I didn't see the `feed_prompt` above the `infer`. (I think the `feed_prompt` can be removed and replaced with `infer` entirely, but might need to test that.) This is fine as-is.

crates/llm-base/src/inference_session.rs
averypelle: Updated!

philpax:
Heya! ...apologies for hijacking the PR. I went to test it and all of your changes worked as expected, but I realised that there were quite a few latent bugs with the stuff not covered by your PR and that the whole chat/REPL logic just wasn't working how I wanted it to work. I ended up revising way more than I intended 😅 

The upshot is that it should now work consistently, and there shouldn't be any surprise discrepancies between REPL and chat mode. Sorry once again for the complete hijack 😭

crates/llm-base/src/inference_session.rs
```diff
@@ -893,32 +893,37 @@ pub fn feed_prompt_callback<'a, E: std::error::Error + 'static>(
...
-/// An [InferenceResponse] callback that will halt inference when a stop_sequence is generated.
+/// An [InferenceResponse] callback that will halt inference when a `stop_sequence` is generated.
 /// This callback is used in [InferenceSession::infer] in chat_mode.
 pub fn conversation_inference_callback<'a, E: std::error::Error + 'static>(
     stop_sequence: String,
-    buf: &'a mut String,
-    mut print_token: impl FnMut(String) + 'a,
+    mut callback: impl FnMut(String) + 'a,
 ) -> impl FnMut(InferenceResponse) -> Result<InferenceFeedback, E> + 'a {
+    let mut stop_sequence_buf = String::new();
```
averypelle, line 904: interesting, scoping this buffer to the function seems a lot better!

crates/llm-base/src/inference_session.rs
```diff
@@ -897,8 +897,8 @@ pub fn feed_prompt_callback<'a, E: std::error::Error + 'static>(
 
 /// An [InferenceResponse] callback that will halt inference when a `stop_sequence` is generated.
 /// This callback is used in [InferenceSession::infer] in chat_mode.
-pub fn conversation_inference_callback<'a, E: std::error::Error + 'static>(
-    stop_sequence: String,
+pub fn conversation_inference_callback<'a, E: std::error::Error + Send + Sync + 'static>(
```
averypelle, line 900: what do these do?

utils/prompts/pygmalion-prelude.txt
```diff
@@ -1,4 +1,3 @@
 Assistant's Persona: Assistant is a highly intelligent language model trained to comply with user requests.
 <START>
-Assistant: How may I help you?
-You: {{PROMPT}}
\ No newline at end of file
+Assistant: How may I help you?
```
averypelle, line 3: I have some prompts I made for Falcon and MPT too since I was testing that. Want me to add in a follow-up PR?

binaries/llm-cli/src/main.rs
```diff
@@ -210,184 +197,12 @@ fn prompt_tokens(args: &cli_args::PromptTokens) -> Result<()> {
...
-    interactive(
-        &args.generate,
-        &args.model_load,
-        true,
-        Some(std::fs::read_to_string(&args.prelude_prompt_file)?.as_str()),
-        Some(&args.message_prompt_prefix()?),
-    )
-}
-
-fn interactive(
```
averypelle, line 242: ah love that you split these out into separate functions!

crates/llm-base/src/inference_session.rs
philpax: In Rust, objects get the `Send` trait if they can be sent across threads, and `Sync` if they can be used by multiple threads (you can see more details [here](https://doc.rust-lang.org/nomicon/send-and-sync.html)). 

I needed to add this because `eyre`, which we use for error reporting in the CLI, expects the error from `infer` to be `Send + Sync`. The error is passed down from `callback` to `infer`, so the trait requirements need to be updated across the library.

utils/prompts/pygmalion-prelude.txt
philpax: Sure thing!

Issue #293, opened 2023-06-02
Support Falcon


zcourts:
Similar to MPT, Falcon is Apache licensed, weights and all!

1. https://huggingface.co/tiiuae/falcon-40b
2. https://huggingface.co/tiiuae/falcon-40b-instruct

And according to the [HuggingFace leaderboard ](https://huggingface.co/spaces/HuggingFaceH4/open_llm_leaderboard) it outperforms all current open source models including MPT.

It seems having a GGML conversion done of the model is a necessary precursor to having [it included](https://github.com/rustformers/llm/pull/218).

I don't think I have the expertise to do this but we may be able to help (e.g. can give access to a V100S or V100S to do the conversion)

zcourts:
Ongoing discussion worth tracking here to get GG conversion https://github.com/ggerganov/llama.cpp/issues/1602 

Found after posting this here. An attempt to convert has been made https://github.com/ggerganov/llama.cpp/issues/1602#issuecomment-1570827592

zcourts:
Looks like our posts overlapped! Great to hear, I've offered to provide GPU access to further the work being done in https://github.com/ggerganov/llama.cpp/issues/1602 - will follow up as that progresses

KerfuffleV2:
There is now a working GGML example for 40B: https://github.com/ggerganov/ggml/pull/231

LLukas22:
That's great! Maybe i will create a draft, but i would like to wait until it get's merged into ggml.

LLukas22:
Yeah, I noticed that. It would be great if someone could try porting it to Rust. I'm currently quite busy implementing GPU acceleration for all architectures.😬

philpax:
Damn, was hoping editing the description would cancel out the issue-closing.

Anyhow - I've merged in the Falcon 7B implementation, but it doesn't handle 40B, and it requires 32-bit memory tensors as the `repeat` operation it uses doesn't work with 16-bit tensors.  Because of these caveats - and the continuing work on (one of) the original implementations in https://github.com/cmp-nct/ggllm.cpp - I've decided to merge it in, but disable it by default.

I'll keep this issue open until Falcon is truly ready to fly.

philpax:
@LLukas22 should we close this or wait until the model format has stabilised?

LLukas22:
We should wait until GGUF is implemented and we have all the necessary fields in the model file.

PR #368, opened 2023-07-15
add Falcon 40B model support

skirodev:
This PR has already passed the tests of [the Falcon mini](https://huggingface.co/Skiro/falcon-mini-ggml) series models, but due to limitations of my device, I haven't tested it with the original Falcon series models. #293


philpax:
Looks promising, I'll leave it to Lukas to check the specifics.

Is the model definition/file format beginning to stabilise a little? Part of the reason it's still experimental is because, as far as I can tell, upstream is still figuring out the specifics of the implementation.

skirodev:
Yeah, perhaps it should remain in an experimental state until it undergoes all necessary tests and stabilizes.

LLukas22:
@skirodev On a first glance this looks good. How did you create your ggmlv3 versions of the model? I would like to try 7B- and 40B-instruct before i merge this.

skirodev:
@LLukas22 I just uploaded the conversion code (only for f16 and f32) to [Hugging Face](https://huggingface.co/Skiro/falcon-mini-ggml). The quantization code is not completed yet.

LLukas22:
Thanks! You an probably just use `llm` to quantize your converted models, it should support the normal Q-quants if you implemented the hyperparams correctly.

skirodev:
Thanks for the suggestion! I'll try using llm to quantize the converted models and see how it works.

skirodev:
The performance of using the falcon-7b-instruct-ggml-q4_0.bin model is shown in the following:

```
cargo run --release -- infer --no-float16 -n 256 -a falcon -m "./models/falcon-7b-instruct-ggmlv3-q4_0.bin" --batch-size 512 -p "write a story about falcon" --stats
```

falcon-7b-instruct-ggml-q4_0

LLukas22:
Alright, I used your script to create ggml versions of falcon-7B and falcon-40B and quantized them with `llm` to `q4_0`.

7B works as expected. Output:
```
write a story about falconry
About
Portfolio
Contacts
For all that is left in the forest, hunt down foxes, wolves, rabbits, and even birds of prey. The first is the hunt, the second is the capture of the bird, and the third is the training of the hawk. Falconers also breed hawks, and some birds may pass through more than one set of hands before they are considered ready for sale or use as a pet. However, you need to be an experienced bird owner before you start a hunt. There are three main types of hunting hawks: lure, live lure, and lure/live. Falconers are trained in many different types of birds, each with its own set of specialized handling techniques and training methods. Falconers hunt with two birds - one as a hawk and the other as a lure bird. For example, falconers use a lure hawk to hunt with live birds and lure hawks to hunt with lure birds. The lure hawk uses birds that are already wild and free to hunt for its own food, whereas the live hawk hunts live birds that are kept in captivity, either for the purpose of training or for hunting purposes. The birds that hunt with lure birds hunt birds like pigeons, doves, and prairie
``` 

Sadly 40B produces gibberish 😞. Output:
```
write a story about falcon/ it or  and  I your body, on the all of 
 not of the  it  new/  can a all you  the so  so. more- this. our past- so in: so I to so you  we'. in  we a  you  it a no.’ your people should a I of it,  by and, and  all the I  to, of

 that what is not all a. for  I can new
 for  in an and we you from all:  you may in of  all
/ to  - D
```

I don't know if the inference code is the problem or if the conversion-script/quantization corrupted some tensors.

crates/ggml/src/context.rs
```diff
@@ -473,6 +473,20 @@ impl Context {
...
+    pub fn op_flash_attn(&self, q: &Tensor, k: &Tensor, v: &Tensor, masked: bool) -> Tensor {
+        let tensor = unsafe {
+            sys::ggml_flash_attn(
+                self.ptr.as_ptr(),
+                q.ptr.as_ptr(),
+                k.ptr.as_ptr(),
+                v.ptr.as_ptr(),
+                masked,
+            )
+        };
```
LLukas22, line 487: I don't know if we should use ggmls flash-attention. The ggml version of the main branch is quite old (~3 weeks) mybe this could lead to errors 🤔

crates/llm/Cargo.toml
```diff
@@ -33,7 +33,7 @@ default = ["models", "tokenizers-remote"]
 
 tokenizers-remote = ["llm-base/tokenizers-remote"]
 
-models = ["llama", "gpt2", "gptj", "bloom", "gptneox", "mpt"]
+models = ["llama", "gpt2", "gptj", "bloom", "gptneox", "mpt", "falcon"]
```
LLukas22, line 36: We should still keep falcon behind a feature flag, until the problem with the linearly scaling performance of the repeat operation is fixed.

crates/models/falcon/src/lib.rs
```diff
@@ -163,18 +183,35 @@ impl KnownModel for Falcon {
...
+                            &current,
+                        ),
+                    );
+                }
+
                 // compute QKV
                 current = ctx0.op_mul_mat(&self.layers[il].query_key_value, &current);
 
-                let fused_qkv_row_nb = (n_embd + 2 * (n_embd / n_head)) * f32_size;
+                let fused_qkv_row_nb = head_dim * (n_head + 2 * n_head_kv) * f32_size;
```
LLukas22, line 214: Is this from the pytorch implementation? Or where did you reference this from?

crates/ggml/src/context.rs
skirodev: I noticed that the upstream already supports broadcast `mut_mat`, perhaps we can use it to replace `repeat`.

crates/llm/Cargo.toml
skirodev: I agree with your opinion.

crates/models/falcon/src/lib.rs
skirodev: The code references [ggllm.cpp](https://github.com/cmp-nct/ggllm.cpp/blob/60f82cac75f4f3be4a85dba397c0632a28de28ba/libfalcon.cpp#L2195), and it should be equal to the value of `current.get_nb()[1]`.

crates/ggml/src/context.rs
LLukas22: We could wait for the CUDA PR to get merged as it is already based on the newest `ggml` version, then we could try to use the new `mut_mat`. What do you think?

skirodev:
> Sadly 40B produces gibberish 😞. Output:
> ```
> write a story about falcon/ it or  and  I your body, on the all of 
>  not of the  it  new/  can a all you  the so  so. more- this. our past- so in: so I to so you  we'. in  we a  you  it a no.’ your people > should a I of it,  by and, and  all the I  to, of
> 
>  that what is not all a. for  I can new
>  for  in an and we you from all:  you may in of  all
> / to  - D
> ```
> I don't know if the inference code is the problem or if the conversion-script/quantization corrupted some tensors.

My apologies, could you retest it now?

crates/ggml/src/context.rs
skirodev: Sounds good.

LLukas22:
```
     Running `target\release\llm.exe infer --no-float16 -n 256 -a falcon -m C:\Users\lkreu\Desktop\falcon\falcon-40b-q4_0 --batch-size 512 -p "write a story about falcon" --stats -r tiiuae/falcon-40b`
⣟ Loading model...[2023-07-16T10:54:11Z INFO  cached_path::cache] Cached version of https://huggingface.co/tiiuae/falcon-40b/resolve/main/tokenizer.json is up-to-date
✓ Loaded 484 tensors (23.5 GB) after 750ms
write a story about falcon, our you this the . more to  not and
 on your so the- you.  I. it. we may a so it for that or any
 the and, and for more the and  we  I by no  so in and just   of our- a can of. all,? I it  at the 
 so,  is it of my personal to that. 
 the and by is it
```

It got a bit better but there probably still is something wrong. Maybe we should wait for the ggml update and revisit it then?

skirodev:
> It got a bit better but there probably still is something wrong. Maybe we should wait for the ggml update and revisit it then?

Absolutely. I agree with your suggestion. Let's await the ggml update and take another look at it when the time comes. Thank you for conducting the testing.

LLukas22:
Could you try to pull the latest main into this? It should contain the latest ggml version.

skirodev:
> Could you try to pull the latest main into this? It should contain the latest ggml version.
Sure, I already pulled the latest main branch for the updated ggml version.

LLukas22:
Falcon 40B still produces gibberish: 
```
PS F:\Github\llm-main> cargo run --release --features falcon -- infer -a falcon -p "Tell me a story about a falcon" -m "C:\Users\lkreu\Desktop\falcon\falcon-40b-q4_0" --no-float16
    Finished release [optimized] target(s) in 0.24s
     Running `target\release\llm.exe infer -a falcon -p "Tell me a story about a falcon" -m C:\Users\lkreu\Desktop\falcon\falcon-40b-q4_0 --no-float16`
✓ Loaded 484 tensors (23.5 GB) after 114ms
Tell me a story about a falcon . more to or- they in, you your. de/ /   to  they.: 
  you, I so good more on to one. on a for all  I  so,.  i  or  and - and it  that  :: and la
```

Maybe i need to reconvert/quantize my model 🤔

skirodev:
[Falcon 40B](https://huggingface.co/Skiro/falcon-40b-instruct-ggml) is now capable of successful inference after my testing. 

```bash
cargo run --release -- infer -a falcon -m "./models/falcon-40b-instruct-ggmlv3-q4_0.bin" --batch-size 512 -p "write a story about falcon" -r tiiuae/falcon-40b --stats

The majestic bird of prey soared through the sky, its wingspan stretching outwards as it searched for prey. Its sharp 
eyes scanned the horizon, and in an instant, it spotted movement below. With powerful strokes of its wings, it dove 
towards its target at incredible speeds before striking with lightning-fast precision. The falcon was a symbol of 
strength, agility, and intelligence – an awe-inspiring creature that commanded respect from all who saw it soar above.
```

However, the embedded tokenizer code still needs modification as Falcon does not require adding bos token id and has some special tokens, which may depend on  the implementation of [GGUF format.](https://github.com/ggerganov/llama.cpp/pull/2398)

LLukas22:
Good Job 👍 

I'll give this another look tomorrow and if everything works i'm gonna merge it.

philpax (approved):
Code looks good! Will leave it to Lukas to do final tests but OK from my end. Hopefully we can get the additional information from GGUF soon.

LLukas22:
LGTM 👍 

Now even works with the fp16 memory :D

LLukas22 (approved):

LLukas22:
@jempabroni Maybe, depends on if all necessary operations were already ported into metal shaders. You can try using it and if it gives you an invalid operation error it's not supported yet.

PR #370, opened 2023-07-16
feat(model/clone): derive `Clone` for `Box<dyn Model>`

radu-matei:
This builds on top of the work @louisgv did in #263.

This commit addresses the following scenario: for a service that is supposed to start multiple inference sessions based on the same model, the performance of loading the model on every request is (at least based on empirical evidence) not consistent.

This commit makes it possible to clone a model and create a new inferencing session without paying the cost of loading the model on every request.

The implementation ultimately implements `Clone` for `Tensor` by calling the `share` function.
The initial attempt does this because it reduces the `Clone` implementations for each model to deriving it, rather than having an explicit `impl` — the alternative would mean manually calling `share` for every layer and every tensor for the implemented models — happy to take this approach is it makes more sense.

### DIFF
--- a/crates/ggml/src/tensor.rs
+++ b/crates/ggml/src/tensor.rs
@@ -9,6 +9,12 @@ pub struct Tensor {
     pub(crate) ctx: Weak<NonNull<sys::ggml_context>>,
 }
 
+impl Clone for Tensor {
+    fn clone(&self) -> Self {
+        self.share()
+    }
+}
+
 impl Tensor {
     /// Size of the `ggml_tensor` struct in bytes.
     ///
--- a/binaries/llm-test/src/main.rs
+++ b/binaries/llm-test/src/main.rs
@@ -196,7 +196,9 @@ async fn test_model(
         local_path: &'a Path,
     }
     impl<'a> llm::ModelArchitectureVisitor<anyhow::Result<()>> for TestVisitor<'a> {
-        fn visit<M: llm::KnownModel + 'static>(&mut self) -> anyhow::Result<()> {
+        fn visit<M: llm::KnownModel + 'static + std::clone::Clone>(
+            &mut self,
+        ) -> anyhow::Result<()> {
             let Self {
                 model_config,
                 test_config,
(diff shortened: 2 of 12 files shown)


crates/ggml/src/tensor.rs
```diff
@@ -9,6 +9,12 @@ pub struct Tensor {
     pub(crate) ctx: Weak<NonNull<sys::ggml_context>>,
 }
 
+impl Clone for Tensor {
```
philpax, line 12: I was considering doing this myself, but decided not to do it because the semantics are unclear - that is, you are not actually cloning the tensor data, just a handle to it, which means you can't take your newly cloned `Tensor` away from the context that owns it.

Maybe we should rename this to `TensorHandle` or something?

philpax:
Hmmm... I'm not sure about this. I think this will not generally do what users expect - it won't create an entirely new instance of the model, but will instead create another handle to the same resource. 

My plan for `Model` was to make `Clone` a deep clone, so that you would have an entirely independent model (where a sufficiently-configured system could do parallel inferencing without accessing the same memory).

Would `Arc::from_box` work for your application? A quick test shows that 
```rust
    let model: Box<dyn llm::Model> = args.model_load.load(args.generate.use_gpu)?;
    let model: Arc<dyn llm::Model> = Arc::from(model);
    let model2: Arc<dyn llm::Model> = model.clone();
```
seems to work as expected, and should mirror the semantics here.

crates/ggml/src/tensor.rs
radu-matei: If we decide not to implement `Clone` for `Tensor` this way, we can still achieve the same by manually implementing `Clone` for each model's `Layer` by calling `.share()` on each of its tensors.

That would be more verbose on each model implementation, but would make it clear that `Tensor` doesn't implement `Clone`.

What would you suggest renaming to `TensorHandle`?

LLukas22:
The problem i see with this is that if the source model disposes it's context all "clones" you created will panic if you want to infer with them. Which could be very confusing for a user 🤔

radu-matei:
Yeah, I think you both make a good argument against this.

The solution @philpax suggests should solve this scenario. Closing this, but creating a deep clone for a tensor would still be helpful in the future.

Thanks, both!

PR #376, opened 2023-07-17
Peg CUDA architecture

LLukas22:
Fixes https://github.com/rustformers/llm/issues/373


philpax (approved):
Looks good. Before I merge, can you confirm you haven't seen any instances of the bug with all of the models you've tested? (just making sure we aren't closing the issue before confirming this was the cause)

LLukas22:
Well i first updated `llama.cpp` which resulted in all models, even the ones that previously worked, to produce garbled output. After i defined the architecture the models work fine, but i will run most of my local models through it just to be sure.

LLukas22:
Alright i tested this against a mix of 3B, 7B ,13B and 30B models with `q4_0`, `q5_1`, `q4_k` and `q5_k` quantizations. Everything seams to work fine 👍

PR #400, opened 2023-08-09
fix(deps): remove pinned `half` dependency

radu-matei:
This commit removes the pinned version of the `half` crate, allowing consumers to resolve to a version of `half` that is compatible with other dependencies in the project.

I'm not sure if pinning to this exact version was intended — happy to close if this causes unintended behavior.

Thanks!

### DIFF
--- a/crates/llm-base/Cargo.toml
+++ b/crates/llm-base/Cargo.toml
@@ -21,7 +21,7 @@ thiserror = { workspace = true }
 partial_sort = "0.2.0"
 serde_bytes = "0.11"
 memmap2 = { workspace = true }
-half = "=2.2.1"
+half = "2"
 tokenizers = {version="0.13.3", default-features=false, features=["onig"]}
 regex = "1.8"
 tracing = { workspace = true }


LLukas22:
I think there once was an error with newer versions of `half` only building on rust >= 1.70 which was a problem, butt since the CI runs without problems it should be fine to remove the pin 👍

philpax:
Sorry, I've been very busy and didn't have time to get around to this...

I *think* this should work, but the problem is if `half` gets resolved with `2.3`, which [raised the MSRV](https://github.com/starkat99/half-rs/issues/97):

```
llm # cargo update -p half
    Updating crates.io index
    Updating half v2.2.1 -> v2.3.1
llm # cargo +1.65 run
    Updating crates.io index
  Downloaded llm-samplers v0.0.6
  Downloaded 1 crate (26.3 KB) in 0.71s
error: package `half v2.3.1` cannot be built because it requires rustc 1.70 or newer, while the currently active rustc version is 1.65.0
Either upgrade to rustc 1.70 or newer, or use
cargo update -p half@2.3.1 --precise ver
where `ver` is the latest version of `half` supporting rustc 1.65.0
```

I don't think it should be an issue because the CLI should resolve with the 2.2.1 lock, but I'm not sure about library use. We'll find out, I guess! 😅

PR #401, opened 2023-08-10
Add the ability to quantize into k-quants

LLukas22:
Closes https://github.com/rustformers/llm/issues/330.

Currently this doesn't implement the "upgrade" logic of the k-quants where certain layers get quantized in a higher level depending on the postfix (`S`,`M`,`L` ) of the quantization. Due to this `q2_k` and `q3_k` are currently not usable/produce gibberish.

### DIFF
--- a/crates/llm-base/src/quantize.rs
+++ b/crates/llm-base/src/quantize.rs
@@ -9,6 +9,7 @@ use half::f16;
 use regex::Regex;
 use std::{
     collections::HashMap,
+    fmt,
     io::{BufRead, Seek, Write},
     path::PathBuf,
     sync::Arc,
@@ -36,6 +37,17 @@ pub enum QuantizeProgress<'a> {
     TensorQuantizing {
         /// Name of the tensor.
         name: &'a str,
+    },
+    /// A tensor is being quantized.
+    TensorFallback {
+        /// Name of the tensor.
+        name: &'a str,
+        /// Size of the tensor.
+        dims: [usize; 2],
+        /// Quantization target.
+        target: QuantizationTarget,
+        /// Quantization fallback.
+        fallback: QuantizationTarget,
     },
     /// A tensor has been quantized.
     TensorQuantized {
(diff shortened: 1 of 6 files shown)


crates/llm-base/src/quantize.rs
```diff
@@ -37,6 +38,17 @@ pub enum QuantizeProgress<'a> {
         /// Name of the tensor.
         name: &'a str,
     },
+    /// A tensor is being quantized.
```
philpax, line 41: Update this doc comment to explain how it's different from `TensorLoading`

philpax:
Seems pretty reasonable. Not sure how we should call out the lack of upgrading - maybe we just mention it in all of the doc comments and CLI options, and try to figure out a solution for it at some point.

Let me know what's left to do here.

LLukas22:
Currently i just thought about adding some sort of `sensitive layer` regex to each architecture which defines which layers should be upgraded. The i would extend the quantiz command to accept a "quantization suffix" (`s`,`m`,`l`) which decides how much these layers will be "upgraded" which should work similar to this: 

1. `s` => no upgrades
2. `m` => +1
3. `l` => +2

This should result in very similar results to `llama.cpp` while being generalizable to many architectures. The only thing that is still missing after this is the logic that some layers only get upgraded after a certain number of attention blocks, meaning we would need a way to determine that. What are your thoughts about this?

philpax:
Seems pretty reasonable - we already have something like that in place for the quantization itself. It might be a bit finicky without standardised names, but it looks like they're working on that as part of GGUF, too. Feel free to do it!

As for the attention blocks thing... uh... we could do something similar and specify that as a model parameter, too. Actually, maybe we should just have it return a `QuantizationSettings` (suffix pending, not sure if that conveys the right meaning) that controls which tensors are quantized and how they're upgraded?

LLukas22:
Maybe we could create some sort of `Quantizable` trait for our models which exposes function that handle which layers get quantized/upgraded etc. That way we could have architectures that can't be quantized in the future e.g. BERT. But I'm a bit unsure about that proposal. I personally would like to wait and see what quantization format candle will implement. I don't like the current state with 10+ quantization options as it can be a bit confusing for new users.

LLukas22:
Closing this for now, as i maybe look into solving this via candle.

Issue #402, opened 2023-08-11
Supporting LLama-2 70B param


AmineDiro:
Hello, 
First of all, I want to thank you for this amazing crate! It is truly a joy to work on LLM using Rust 😄 .

I recently wrote an [API that serves Llama-2](https://github.com/AmineDiro/cria) models using this crate. 

I have an issue for serving Llama2-70B-GGML model. The 65B llama and 70B Llama-2 models use grouped query attention. This is done in `llama.cpp` by specifying the `n_gqa` params in model hyperparameters which feels a little bit hacky 🤔  

 I would love to work on adding support for the `n_gqa` on this crate, I think that it can be added to the Llama model: 
```rust
/// LLaMA [hyperparameters](https://en.wikipedia.org/wiki/Hyperparameter_(machine_learning))
#[derive(Debug, Default, PartialEq, Eq, Clone, Copy)]
pub struct Hyperparameters {
  ....
   pub n_head_kv: usize,
}
````
Where the n_head_kv is   `n_head_kv = n_head / n_gqa;` and `n_head % n_gqa == 0` and pass the n_gqa parameter in `ModelParameters` as Optional 🤔 

Thank you for you help ! 
`

LLukas22:
If you want you could create a PR containing these changes. The only thing i don't like about it is that the `ModelParameters` will then contain the `n_gqa` parameter which is only used by llama models and no other architectures but i guess until GGUF is finalized this is the best way to implement it 🤔

philpax:
Yes, I've been considering implementing this but have held off because of the need to parameterise the model just for something that should hopefully be removed in a few weeks.

We actually had [a mechanism for this before for this exact purpose](https://github.com/rustformers/llm/pull/204/commits/08bc9f9b45275ead6efa8390cb6ab60bee5791d7), but I [removed it](https://github.com/rustformers/llm/commit/8f62b39ade8d4282077995a2f6d0b6f9057d301d) after the model format was updated with a breaking change to remove the specific parameter.

I don't think I want to reintroduce it again - especially as we should never need to do this again - but I would like to support 70B at some point. I'm not sure. Maybe a bodge function on the `Llama` model that lets you set `n_head_kv` after the fact?

AmineDiro:
> I don't think I want to reintroduce it again - especially as we should never need to do this again - but I would like to support 70B at some point. I'm not sure. Maybe a bodge function on the `Llama` model that lets you set `n_head_kv` after the fact?

[...]

PR #409, opened 2023-08-17
Support for Llama 70-B

AmineDiro:
Hello, 
Solves #402 .

This is a temporary fix for supporting the Llama-2 70B model. I wanted to open a draft PR to get your feedbacks on this implementation for supporting the `n_gqa` params : 
- Added `n_gqa` as a optional param in `ModelParameters`
- Added `LlamaModelVersion` enum akin to other [e_model union in llama.cpp](https://github.com/ggerganov/llama.cpp/blob/8dae7ce68437faf1fa96ec0e7687b8700956ef20/llama.cpp#L76)
- Modified self-attention evaluation to use the `n_head_kv` for `K` and `V` instead of `n_head`

Here is the `llama-2-70B--chat.ggmlv3.q4_0.bin` model loaded on A100 GPU : 
Annotation 2023-08-17 203513


crates/ggml/sys/build.rs
```diff
@@ -332,6 +332,7 @@ fn enable_cublas(build: &mut cc::Build, out_dir: &Path) {
             .arg("static")
             .arg("--generate-code=arch=compute_52,code=[compute_52,sm_52]")
             .arg("--generate-code=arch=compute_61,code=[compute_61,sm_61]")
+            .arg("--generate-code=arch=compute_75,code=[compute_75,sm_75]")
```
LLukas22, line 335: Is this needed for newer cards?

crates/models/llama/src/lib.rs
```diff
@@ -447,3 +486,13 @@ struct Layer {
...
+
+/// Available Llama models
+enum LlamaModelVersion {
+    Model3b,
+    Model7b,
+    Model13b,
+    Model30b,
+    Model65b,
+    Model70b,
+}
```
LLukas22, line 498: Version hints at the actual llama model version (1 or 2), maybe rename it into `LlamaModelType`?

LLukas22:
Looks good, some small nitpicks but if the CI passes it should be good to go 👍

crates/ggml/sys/build.rs
AmineDiro: Yes  I had issues with my card RTX2070, I guess we should also support 8.x architectures like A100

crates/models/llama/src/lib.rs
AmineDiro: That's true, I changed this in [a16caba] (https://github.com/rustformers/llm/pull/409/commits/a16cabaf5b9b5c1711fe4689e3c00f492524f4e7)

AmineDiro:
@LLukas22  Thanks for the review 👍🏼 !

LLukas22:
Thanks for implementing this :D

PR #410, opened 2023-08-18
add `Tensor::is_contiguous`

nerdypepper:
`Tensor::is_contiguous`: returns true if this tensor is stored contiguously in memory


LLukas22:
Yup, good to have this addition, otherwise it's hard to tell when a matmul can be performed 👍

PR #412, opened 2023-08-20
GGUF support

### LINKED ISSUE #365: Support GGUF
GGUF is the new file format specification that we've been designing that's designed to solve the problem of not being able to identify a model. The specification is here: https://github.com/ggerganov/ggml/pull/302

`llm` should be able to do the following:
- continue supporting existing models (i.e. this change should be non-destructive)
- load GGUF models and automatically dispatch to the correct model. 
  - `load_dynamic` already has an interface that should support this, but loading currently only begins *after* the model arch is known
  - use the new information stored within the metadata to improve the UX, including automatically using the HF tokenizer if available
- save GGUF models, especially in quantization

`llm` could do the following:
- convert old models to GGUF models with prompting for missing data
- implement the migration tool mentioned in the spec, which does autonomous conversion for users based on hashes

philpax:
Implements support for loading and saving GGUF support.

TODO:
- [x] Implement a basic GGUF format loader that loads the file in as a struct.
- [x] Load a Llama model with a remote Hugging Face tokenizer.
  - [ ] Load a Llama model with an embedded tokenizer (GGML).
  - [x] Load a Llama model with an embedded tokenizer (HF).
  - [x] Ensure that models with a HF tokenizer can operate without a GGML tokenizer.
- [x] Implement a basic GGUF format saver that saves a `Gguf` struct to a file.
  - [ ] Reimplement `quantize`. (For extra points, make it multithreaded.)
  - [x] Think about how hyperparameters should generally be treated. Should a model re-write them to a `Metadata` map?
- [x] Make sure all `llm` metadata values are used for `llama`.
- [ ] Fix all the models.
  - [ ] Fix BLOOM.
  - [ ] Fix Falcon.
  - [ ] Fix GPT-2.
  - [ ] Fix GPT-J.
  - [ ] Fix GPT-NeoX.
  - [ ] Fix MPT.
- [ ] Remove all of the `expect`s.
- [x] Remove the `architecture` option and load entirely based on the architecture specified in the GGUF.

Open questions:
- Should we still support the old formats?
  - No. Instead, we'll take the old code and build a converter with it. Preferably one that can ingest HF files to provide the necessary information for a fully-compliant GGUF. 
- How resilient should we be to malformed GGUF models?
  - Answer: The usual Rust standard. Don't panic if you can avoid it.

Closes #365.


svenstaro:
I think having a migration tool for converting previous formats to GGUF and then removing support for other models might be the most maintainable solution. It might be too early to definitely call this but I think it's prudent to assume that the ecosystem will converge on GGUF as the preferred format soon.

KerfuffleV2:
I've been messing around cleaning up the Python scripts in llama.cpp (like the converters, Python side of GGUF) so if you need to pick someone's brain about GGUF stuff I might be able to help. I'm _not_ a expert by any means.

philpax:
Aye, I noticed you contributed the conversion script upstream; I'll definitely reach out if I have any questions about the specifics there.

PR #419, opened 2023-09-02
Add Default GPU Split and Runtime Configuration

jacohend:
+ Add configuration item InterferenceSessionConfig                                
``` 
gpu_split: 10,
cuda_device: 0,
```
+ Add configuration item ModelParameters
```
 gpu_backend: ggml::accelerator::Backend::Gpu,
````

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "crates/ggml/sys/llama-cpp"]
 	path = crates/ggml/sys/llama-cpp
 	url = https://github.com/ggerganov/llama.cpp
+[submodule "crates/ggml/sys/llama.cpp"]
+	path = crates/ggml/sys/llama.cpp
+	url = git@github.com:jacohend/llama.cpp.git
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: .gitmodules [hunk 1]
```diff
@@ -1,3 +1,6 @@
 [submodule "crates/ggml/sys/llama-cpp"]
 	path = crates/ggml/sys/llama-cpp
 	url = https://github.com/ggerganov/llama.cpp
+[submodule "crates/ggml/sys/llama.cpp"]
+	path = crates/ggml/sys/llama.cpp
+	url = git@github.com:jacohend/llama.cpp.git
```
LLukas22, line 6: Uhm this shouldn't be here right?

binaries/llm-cli/Cargo.toml
```diff
@@ -13,6 +13,7 @@ path = "src/main.rs"
 
 [dependencies]
 llm = { path = "../../crates/llm", version = "0.2.0-dev", default-features = false, features = ["models"] }
+ggml = { path = "../../crates/ggml", version = "0.2.0-dev" }
```
LLukas22, line 16: We probably don't want to expose this.

crates/llm-base/src/inference_session.rs
```diff
@@ -144,7 +144,7 @@ impl InferenceSession {
         };
 
         if use_gpu {
-            ggml::accelerator::initialize(0);
+            ggml::accelerator::initialize(config.cuda_device as i32, config.gpu_split);
```
LLukas22, line 147: Does this work with multiple gpus? I always thought we need to define every device id we want to use.

PR #426, opened 2023-09-14
Fix build on macos m1

olexiyb:
During a build using Jenkins agent I got an error `-mcpu=native` is not available for clang


LLukas22:
Looks good. Just to confirm, this only occurs if you target macos on a non mac machine?

PR #429, opened 2023-09-16
few unwrap removes

chris-ha458:
map_or can be used instead of map()..unwrap_or()


philpax:
It would be nice to use `is_some_and`, but that's not in our MSRV. I'll take this for now.

